Issue #20, opened 2022-07-19
CDN-friendly HTTP API


mildbyte:
Implement a read-only HTTP API that obeys HTTP cache semantics and can benefit from any CDN / cache like Varnish:

Client sends a query like:

```
GET /q/[SHA hash of the query text]
X-Seafowl-Query: [actual query text]
```

The server receives the query, checks it's a SELECT query, checks that the hash matches, executes it. It sends it back with an [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) that is a function of the versions of all tables that participate in this query:

```
200 OK
[query results]
ETag: [hash of the version of the table(s)]
Cache-Control: max-age=3600 [optional]
```

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#directives

When the client's browser queries the data again, it can then pass the ETag to see if the query results have changed:

```
GET /q/[SHA hash of the query text]
X-Seafowl-Query: [actual query text]
If-None-Match: [etag]
```

Intermediate caches/CDNs might not even forward the query to the origin, instead serving it from cache if it's not stale. If it reaches the server, the server can cheaply revalidate the entry by recomputing the etag and responding with a `304 Not Modified` if the tables in the query haven't changed their versions (without having to execute the query).

This only works for `SELECT` queries. `INSERT` and other writes should come in via `POST` requests.

mildbyte:
> We might be able to make the GET API more friendly without losing caching semantics. If the request is GET /q/SELECT%201 (i.e., the parameter is the raw query, not a hash), then we could hash the parameter and return a 301 redirect to GET /q/[hash]

That's a decent idea actually, though the 301 would have to indicate to the requester to not only resubmit to `GET /q/[hash]` but also include the query as a header or GET body. Is there a way to force a 301 / 307 to do that?

milesrichardson:
I'm not aware of a way for the server to provide the client with a _new_ body to send. But a 307 redirect will [reuse the method and body ](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307) of the request, so it's not perfect but it works if the client provides a copy of the query in both the body and URL parameter (or even only in the body).

milesrichardson:
There is [303 See Other](https://httpwg.org/specs/rfc7231.html#status.303) ([wiki](https://en.wikipedia.org/wiki/HTTP_303)) which is interesting but doesn't seem to allow changing the body of the next request. Worth a read though

Issue #11, opened 2022-07-19
Add V0 authn/z


mildbyte:
- unauthenticated requests are read-only
- authenticated requests (with a pre-shared secret) can write to the database

PR #47, opened 2022-08-11
HTTP API implementation

mildbyte:
Some more implementation for https://github.com/splitgraph/seafowl/issues/20 and other fixes:

  - add CORS
  - add optional `.extension` to `GET /q/hash[.extension` (ignored by us, but used by Cloudflare to see if it's a cacheable asset)
  - add support for query in the `GET` body
  - add `POST /q` to execute uncached RW queries (no authz until https://github.com/splitgraph/seafowl/issues/11)
  - replace mock context with in-memory SQLite (does mean we end up relying on some parts of the context's implementation but is much more concise than using mocks)
  - simplify some routing/parsing

### DIFF
--- a/seafowl.toml
+++ b/seafowl.toml
@@ -5,8 +5,8 @@ type = "local"
 data_dir = "../seafowl-data"
 
 [catalog]
-type = "postgres"
-dsn = "postgresql://sgr:password@localhost:7432/seafowl"
+type = "sqlite"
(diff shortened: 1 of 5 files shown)


mildbyte:
(this also has a copypaste of datafusion's implementation for `CreateExternalTable` which we weren't supporting, will keep on that in a separate PR to add support for different URL schemas (`file://` / `http(s)://`))

Issue #40, opened 2022-08-03
HTTP API: uploading data


mildbyte:
Something like:

```
POST /upload/[schema]/[table]
```

receives a standard multi-part upload so that a user can upload with curl. Considerations:

  - Find a way to specify a schema for CSV files (GET params?)
  - Support uploading Parquet files
  - (stretch goal) support uploading SQLite?

PR #51, opened 2022-08-15
Implement CSV/Parquet upload endpoint

gruuya:
This addresses the issue #40, by providing a POST endpoint `/upload/[schema]/[table]`, whereby the provided file contents are parsed and used to create/append to table.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -875,9 +875,8 @@ impl SeafowlContext for DefaultSeafowlContext {
                 // This is actually CREATE TABLE AS
                 let physical = self.create_physical_plan(&input).await?;
 
-                let _res = self
-                    .execute_plan_to_table(&physical, Some(name), None)
-                    .await;
+                self.execute_plan_to_table(&physical, Some(name), None)
+                    .await?;
 
                 Ok(make_dummy_exec())
             }
@@ -911,13 +910,12 @@ impl SeafowlContext for DefaultSeafowlContext {
                         SeafowlExtensionNode::Insert(Insert { table, input, .. }) => {
                             let physical = self.create_physical_plan(input).await?;
 
-                            let _res = self
-                                .execute_plan_to_table(
-                                    &physical,
-                                    None,
-                                    Some(table.table_version_id),
-                                )
-                                .await;
+                            self.execute_plan_to_table(
+                                &physical,
+                                None,
+                                Some(table.table_version_id),
+                            )
+                            .await?;
 
                             Ok(make_dummy_exec())
                         }
--- a/src/context.rs
+++ b/src/context.rs
@@ -361,6 +361,7 @@ pub trait SeafowlContext: Send + Sync {
     async fn plan_to_table(
         &self,
         plan: Arc<dyn ExecutionPlan>,
+        schema_name: String,
         table_name: String,
     ) -> Result<bool>;
 }
@@ -1014,10 +1015,47 @@ impl SeafowlContext for DefaultSeafowlContext {
     async fn plan_to_table(
         &self,
         plan: Arc<dyn ExecutionPlan>,
+        schema_name: String,
         table_name: String,
     ) -> Result<bool> {
+        let mut full_table_name = Some(format!("{}.{}", schema_name, table_name));
+        let mut from_table_version = None;
+
+        // Ensure the schema exists prior to creating the table
+        match self
+            .table_catalog
+            .get_collection_id_by_name(&self.database, &schema_name)
+            .await
+        {
+            Some(_) => {
+                if let Ok(table) =
+                    self.try_get_seafowl_table(full_table_name.clone().unwrap())
+                {
+                    // Table exists, see if the schemas match
+                    if table.schema.arrow_schema != plan.schema() {
+                        return Err(DataFusionError::Execution(
+                            format!(
+                                "The table {} already exists but has a different schema than the one provided.",
+                                full_table_name.clone().unwrap())
+                            )
+                        );
+                    }
+
+                    // Instead of creating a new table, just insert the data into a new version
+                    // of an existing table
+                    full_table_name = None;
+                    from_table_version = Some(table.table_version_id);
+                }
+            }
+            None => {
+                self.table_catalog
+                    .create_collection(self.database_id, &schema_name)
+                    .await;
+            }
+        };
+
         let res = self
-            .execute_plan_to_table(&plan, Some(table_name), None)
+            .execute_plan_to_table(&plan, full_table_name, from_table_version)
             .await;
 
         res
--- a/src/context.rs
+++ b/src/context.rs
@@ -1054,11 +1054,8 @@ impl SeafowlContext for DefaultSeafowlContext {
             }
         };
 
-        let res = self
-            .execute_plan_to_table(&plan, full_table_name, from_table_version)
-            .await;
-
-        res
+        self.execute_plan_to_table(&plan, full_table_name, from_table_version)
+            .await
     }
 }
 
--- a/src/context.rs
+++ b/src/context.rs
@@ -410,6 +410,11 @@ pub trait SeafowlContext: Send + Sync {
         &self,
         physical_plan: Arc<dyn ExecutionPlan>,
     ) -> Result<Vec<RecordBatch>>;
+    async fn plan_to_table(
+        &self,
+        plan: Arc<dyn ExecutionPlan>,
+        table_name: String,
+    ) -> Result<bool>;
 }
 
 impl DefaultSeafowlContext {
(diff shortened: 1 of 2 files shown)


src/frontend/http.rs
```diff
@@ -64,139 +74,276 @@ struct QueryBody {
...
+                        .with_escape(b'\\'); // default is None, change to \
 
-                    let mut buf = Vec::new();
-                    let mut writer = LineDelimitedWriter::new(&mut buf);
-                    writer.write_batches(&batches).unwrap();
-                    writer.finish().unwrap();
+                    let mut cursor = Cursor::new(&value);
+                    let csv_reader = builder.build(&mut cursor).unwrap();
+                    let partition: Vec<RecordBatch> =
+                        csv_reader.into_iter().map(|item| item.unwrap()).collect();
```
mildbyte, line 244: this avoids the extra map (arguably it's more code, but it lets us postpone the unwrap() until the very end; good for when we actually start propagating errors upwards instead of panicking).

```suggestion
                        csv_reader.into_iter().collect::<Result<Vec<RecordBatch>, ArrowError>>().unwrap();
```

src/frontend/http.rs
```diff
@@ -279,6 +439,57 @@ mod tests {
...
+        table_name: &str,
+        schema_json: Option<&str>,
+        context: Arc<dyn SeafowlContext>,
+    ) -> Response<Bytes> {
+        let handler = filters(context);
+
+        let mut body: Vec<u8> = vec![];
+
+        if let Some(schema_json) = schema_json {
+            body = "--42\r\n\
```
mildbyte, line 456: Is there something more standard here we can use for the multipart/form-data boundary (though it looks like the actual server code doesn't use a hardcoded boundary anyway)?

mildbyte (commented):
I did a first pass and left some comments. I still need to figure out if there's some way to simplify this stream-to-Bytes-to-Reader handling code.

src/frontend/http.rs
```diff
@@ -279,6 +439,57 @@ mod tests {
...
+                Content-Disposition: form-data; name=\"data\"; filename=\"{}\"\n\
+                Content-Type: application/octet-stream\n\n",
+                filename
+            )
+            .as_bytes()
+            .to_vec(),
+        );
+
+        body.append(&mut file_content);
+        body.append(&mut "--42--".as_bytes().to_vec());
```
mildbyte, line 478: Can you also test an upload that actually has multiple parts (i.e. more than one boundary)? It could be easier to test this with a test in `end_to_end.rs` that calls out into `curl` to make sure this works with real clients.

src/context.rs
```diff
@@ -874,39 +911,12 @@ impl SeafowlContext for DefaultSeafowlContext {
...
-                                .await;
-                            self.partition_catalog
-                                .append_partitions_to_table(partition_ids, new_version_id)
+                            let _res = self
+                                .execute_plan_to_table(
+                                    &physical,
+                                    None,
+                                    Some(table.table_version_id),
+                                )
                                 .await;
```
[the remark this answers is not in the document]
gruuya: Yup, definitely; I rushed through this part and forgot to revise it afterwards.

src/frontend/http.rs
```diff
@@ -64,139 +74,276 @@ struct QueryBody {
...
+                })
+                .unwrap();
+
+            // Parse the schema and load the file contents into a vector of record batches
+            let (schema, partition) = match filename.split('.').last().unwrap() {
+                "csv" => {
+                    let schema = match csv_schema.clone() {
+                        Some(schema) => schema,
+                        None => {
+                            // TODO: We could actually make the schema part optional, and utilize
```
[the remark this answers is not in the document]
gruuya: Agreed, going with schema inference as default now.

src/frontend/http.rs
```diff
@@ -64,139 +74,276 @@ struct QueryBody {
...
                 }
-            },
+                "parquet" => {
+                    let mut parquet_reader =
+                        ParquetFileArrowReader::try_new(Bytes::from(value)).unwrap();
+
+                    let schema = parquet_reader.get_schema().unwrap();
+
+                    let partition: Vec<RecordBatch> = parquet_reader
+                        .get_record_reader(100000) // TODO: Probably a constant or even a config somewhere
```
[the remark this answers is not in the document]
gruuya: Seems like the usual choice is 1024, so I'm going with that.

src/frontend/http.rs
gruuya: In [theory](https://www.rfc-editor.org/rfc/rfc7578#section-4.1) there's no enforcement on the boundary other than it can't appear in the body. In practice I see it's usually something like `------------------------3d5af745b8be129c`

src/frontend/http.rs
```diff
@@ -64,138 +75,279 @@ struct QueryBody {
...
+                .unwrap();
+
+            // Parse the schema and load the file contents into a vector of record batches
+            let (schema, partition) = match filename.split('.').last().unwrap() {
+                "csv" => {
+                    // If the schema part wasn't specified we'll need to infer it
+                    let builder = match csv_schema.clone() {
+                        Some(schema) => ReaderBuilder::new()
+                            .with_schema(Arc::new(schema.clone()))
+                            .has_header(has_header), // .with_escape(b'\\'); // default is None, change to \
```
mildbyte, line 237: Is this `with_escape` comment still needed? 

```suggestion
                            .has_header(has_header),
```

mildbyte, line 237: If in the future we want to let the user to be able to override some other CSV options like the separator, will we want to continue using the formdata for it or should we use some other way? I had this other idea which might be more appropriate in that case: we let the user pass in a `CREATE EXTERNAL TABLE` query that references the freshly uploaded dataset, e.g.

```sql
CREATE EXTERNAL TABLE my_csv_table ( some: INTEGER, schema: INTEGER)
STORED AS CSV
WITH HEADER ROW, DELIMITER '\t'
LOCATION "@data"
```

similar to Snowflake stages (https://docs.snowflake.com/en/user-guide/querying-stage.html) but it needs some more thinking/API design.

mildbyte (commented):
Looking good! I left some more suggestions on refactoring to avoid some `unwrap()s`.

When you're happy with the API interface, can you do a quick run through https://www.splitgraph.com/docs/seafowl/guides/uploading-csv-parquet and https://www.splitgraph.com/docs/seafowl/reference/http-api#post-uploadschema_nametable_name and make sure it matches the actual API?

(edit: the docs live in https://github.com/splitgraph/marketing/tree/master/content/docs/seafowl)

src/frontend/http.rs
```diff
@@ -488,4 +755,262 @@ SELECT
...
+
+        let has_headers = if let Some(headers) = add_headers {
+            if headers {
+                Some("true")
+            } else {
+                Some("false")
+            }
+        } else {
+            None
+        };
```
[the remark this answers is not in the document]
gruuya: Nice, I missed that.

src/frontend/http.rs
```diff
@@ -64,138 +75,279 @@ struct QueryBody {
...
+                })
+                .await
+                .map_err(|e| {
+                    eprintln!("Error reading part's data: {}", e);
+                    warp::reject::reject()
+                })
+                .unwrap();
+
+            // Parse the schema and load the file contents into a vector of record batches
+            let (schema, partition) = match filename.split('.').last().unwrap() {
```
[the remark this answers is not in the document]
gruuya: Yes, this corresponds to the idea of a single partition. Note that it's not despite, but because of having multiple record batches is this a partition (a single record batch can also make a partition, but by placing it in a `vec!`).

The `MemoryExec` node takes a vector of partitions
```rust
pub struct MemoryExec {
    /// The partitions to query
    partitions: Vec<Vec<RecordBatch>>
...
```
and then [executes](pub struct MemoryExec {
    /// The partitions to query
    partitions: Vec<Vec<RecordBatch>>) by streaming record batches from the partition that is specified.

src/frontend/http.rs
gruuya: I don't think so, removing.

gruuya: Yup, I would say that looks like a suitable use case for formdata as well (as in pass the query in another part of the multipart request with e.g. `-F "query=...").

gruuya:
> Looking good! I left some more suggestions on refactoring to avoid some `unwrap()s`.
> 
> When you're happy with the API interface, can you do a quick run through https://www.splitgraph.com/docs/seafowl/guides/uploading-csv-parquet and https://www.splitgraph.com/docs/seafowl/reference/http-api#post-uploadschema_nametable_name and make sure it matches the actual API?
> 
> (edit: the docs live in https://github.com/splitgraph/marketing/tree/master/content/docs/seafowl)

Certainly, I was thinking of doing that once the PR is merged.

src/frontend/http.rs
```diff
@@ -304,6 +245,60 @@ pub async fn upload(
...
+    let mut parquet_reader =
+        ParquetFileArrowReader::try_new(Bytes::from(source)).unwrap();
+
+    let schema = parquet_reader.get_schema().unwrap();
+
+    let partition: Vec<RecordBatch> = parquet_reader
+        .get_record_reader(1024)
+        .unwrap()
+        .collect::<Result<Vec<RecordBatch>, ArrowError>>()
+        .unwrap();
```
mildbyte, line 297: Can use `?` now since you're returning an ArrowError and it'll propagate it, might be possible to even avoid the turbofish since you have the type specified in the function signature:

```suggestion
    let mut parquet_reader =
        ParquetFileArrowReader::try_new(Bytes::from(source))?;

    let schema = parquet_reader.get_schema()?;

    let partition: Vec<RecordBatch> = parquet_reader
        .get_record_reader(1024)?
        .collect?;
```

src/frontend/http.rs
```diff
@@ -304,6 +245,60 @@ pub async fn upload(
...
+
+    let mut cursor = Cursor::new(source);
+    let csv_reader = builder.build(&mut cursor).unwrap();
+    let schema = csv_reader.schema().as_ref().clone();
+    let partition: Vec<RecordBatch> = csv_reader
+        .into_iter()
+        .collect::<Result<Vec<RecordBatch>, ArrowError>>()
+        .unwrap();
+
+    Ok((schema, partition))
```
mildbyte, line 284:
```suggestion
    let mut cursor = Cursor::new(source);
    let csv_reader = builder.build(&mut cursor)?;
    let schema = csv_reader.schema().as_ref().clone();
    let partition: Vec<RecordBatch> = csv_reader
        .into_iter()
        .collect()?;

    Ok((schema, partition))
```

mildbyte (approved):

PR #60, opened 2022-08-18
Make upload endpoint tests be fully end-to-end

gruuya:
Migrate existing upload tests, and make them use curl instead of mocking the multipart formdata requests by hand.


tests/end_to_end.rs
```diff
@@ -1,3 +1,5 @@
+mod http;
+
```
mildbyte, line 2: Is this actually needed? The tests in `http.rs` ran fine in my case without it.

tests/http.rs
```diff
@@ -162,3 +175,293 @@ async fn test_http_server_reader_writer() {
...
+
+    // Prepare the schema + data (a single record batch) which we'll save to a temp file via
+    // a corresponding writer
+    let schema = Arc::new(Schema::new(vec![
+        Field::new("number", DataType::Int32, false),
+        Field::new("parity", DataType::Utf8, false),
+    ]));
+
+    // For CSV uploads we can supply the schema as another part of the multipart request, to
+    // remove the ambiguity resulting from automatic schema inference
```
mildbyte, line 229: Out of interest, does it correctly infer `number: Int32` (or some numeric-looking value) when the schema is omitted?

mildbyte (approved):
Great! Very nice to see this work end-to-end :)

tests/http.rs
mildbyte: Oh wait, disregard this, it's actually checked in this very test 😅

tests/end_to_end.rs
gruuya: Not quite sure; CLion was complaining that `File is not included in the module tree, analysis is not available` and the intellisense died, until I added it here.

tests/http.rs
gruuya: No it defaults to Int64 for integers, and similarly for other types: https://github.com/apache/arrow-rs/blob/master/arrow/src/csv/reader.rs#L87-L100

PR #67, opened 2022-08-24
Implement min/max/null-count partition pruning

gruuya:
Instead of defaulting to scanning all of the partitions, the goal here is to use the statistical summary for each column in each partition, and based on the filter/qualifier used in the query determine whether some of the partitions are superfluous for getting a correct result. If this determination succeeds simply scope the scan down only to partitions which are relevant to the query.

### DIFF
--- a/migrations/postgres/20220705072248_create_tables.up.sql
+++ b/migrations/postgres/20220705072248_create_tables.up.sql
@@ -43,7 +43,8 @@ CREATE TABLE physical_partition_column (
     name VARCHAR NOT NULL,
     type VARCHAR NOT NULL,
     min_value BYTEA,
-    max_value BYTEA
+    max_value BYTEA,
+    null_count INTEGER
 );
 
 CREATE TABLE table_partition (
--- a/src/context.rs
+++ b/src/context.rs
@@ -161,12 +161,15 @@ fn build_partition_columns(
     // TODO PartitionColumn might not be the right data structure here (lacks ID etc)
     match &partition_stats.column_statistics {
         // NB: Here we may end up with `null_count` being None, but DF pruning algorithm demands that
-        // the null count field be not nullable itself, and consequently for any such cases the
+        // the null count field be not nullable itself. Consequently for any such cases the
         // pruning will fail, and we will default to using all partitions.
         Some(column_statistics) => zip(column_statistics, schema.fields())
             .map(|(stats, column)| {
                 // TODO: the to_string will discard the timezone for Timestamp* values, and will
-                // therefore hinder the ability to recreate them once needed for partition pruning
+                // therefore hinder the ability to recreate them once needed for partition pruning.
+                // However, since DF stats rely on Parquet stats that problem won't come up until
+                // 1) Parquet starts collecting stats for Timestamp* types (`parquet::file::statistics::Statistics` enum)
+                // 2) DF pattern matches those types in `summarize_min_max`.
                 let min_value = stats
                     .min_value
                     .as_ref()
--- a/src/context.rs
+++ b/src/context.rs
@@ -178,7 +178,7 @@ fn build_partition_columns(
                     r#type: Arc::from(column.data_type().to_json().to_string()),
                     min_value: Arc::new(min_value),
                     max_value: Arc::new(max_value),
-                    null_count: stats.null_count.map(|nc| nc as i128),
+                    null_count: stats.null_count.map(|nc| nc as u64),
                 }
             })
             .collect(),
@@ -793,7 +793,7 @@ impl SeafowlContext for DefaultSeafowlContext {
 
                     Ok(LogicalPlan::Extension(Extension {
                         node: Arc::new(SeafowlExtensionNode::Delete(Delete {
-                            name: table_name.to_string(),
+                            name: table_name,
                             selection: selection_expr,
                             output_schema: Arc::new(DFSchema::empty())
                         })),
@@ -1406,7 +1406,8 @@ mod tests {
                             name: Arc::from("timestamp".to_string()),
                             r#type: Arc::from("{\"name\":\"utf8\"}".to_string()),
                             min_value: Arc::new(None),
-                            max_value: Arc::new(None)
+                            max_value: Arc::new(None),
+                            null_count: Some(0),
                         },
                         PartitionColumn {
                             name: Arc::from("integer".to_string()),
@@ -1416,12 +1417,14 @@ mod tests {
                             ),
                             min_value: to_min_max_value(12),
                             max_value: to_min_max_value(42),
+                            null_count: Some(0),
                         },
                         PartitionColumn {
                             name: Arc::from("varchar".to_string()),
                             r#type: Arc::from("{\"name\":\"utf8\"}".to_string()),
                             min_value: Arc::new(None),
-                            max_value: Arc::new(None)
+                            max_value: Arc::new(None),
+                            null_count: Some(0),
                         }
                     ])
                 },
@@ -1433,7 +1436,8 @@ mod tests {
                             name: Arc::from("timestamp".to_string()),
                             r#type: Arc::from("{\"name\":\"utf8\"}".to_string()),
                             min_value: Arc::new(None),
-                            max_value: Arc::new(None)
+                            max_value: Arc::new(None),
+                            null_count: Some(0),
                         },
                         PartitionColumn {
                             name: Arc::from("integer".to_string()),
@@ -1443,12 +1447,14 @@ mod tests {
                             ),
                             min_value: to_min_max_value(22),
                             max_value: to_min_max_value(32),
+                            null_count: Some(0),
                         },
                         PartitionColumn {
                             name: Arc::from("varchar".to_string()),
                             r#type: Arc::from("{\"name\":\"utf8\"}".to_string()),
                             min_value: Arc::new(None),
-                            max_value: Arc::new(None)
+                            max_value: Arc::new(None),
+                            null_count: Some(0),
                         }
                     ])
                 },
@@ -1570,6 +1576,7 @@ mod tests {
                         max_value: to_min_max_value(
                             output_partitions[i].iter().max().unwrap()
                         ),
+                        null_count: Some(0),
                     }])
                 },
             )
@@ -1695,6 +1702,7 @@ mod tests {
                                         r#type: Arc::from("{\"name\":\"date\",\"unit\":\"MILLISECOND\"}"),
                                         min_value: Arc::new(None),
                                         max_value: Arc::new(None),
+                                        null_count: Some(0),
                                     },
                                     PartitionColumn {
                                         name: Arc::from("value"),
@@ -1711,6 +1719,7 @@ mod tests {
                                                 50,
                                             ],
                                         )),
+                                        null_count: Some(0),
                                     },
                                 ],)
                             },]
--- a/src/context.rs
+++ b/src/context.rs
@@ -184,7 +184,7 @@ fn build_partition_columns(
                     r#type: Arc::from(column.data_type().to_json().to_string()),
                     min_value: Arc::new(min_value),
                     max_value: Arc::new(max_value),
-                    null_count: stats.null_count.map(|nc| nc as u64),
+                    null_count: stats.null_count.map(|nc| nc as i32),
                 }
             })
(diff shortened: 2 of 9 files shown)


src/provider.rs
```diff
@@ -174,6 +197,164 @@ impl TableProvider for SeafowlTable {
...
+                }
+            }
+            None => None,
+        }
+    }
+}
+
+impl PruningStatistics for SeafowlPruningStatistics {
+    fn min_values(&self, column: &Column) -> Option<ArrayRef> {
+        self.get_values_array(self.min_values.get(column.name.as_str()))
```
mildbyte, line 339: Why not store the `min_values`/`max_values`/`null_counts` as `Option<ArrayRef>`s to begin with? i.e.: `min_values: HashMap<String, Option<ArrayRef>>`? You're converting them into a Vec anyway in the constructor (and cloning in `get_values_array`, so I don't think this will result in a worse performance.

src/provider.rs
```diff
@@ -174,6 +197,164 @@ impl TableProvider for SeafowlTable {
...
+                            .zip(expr_mask)
+                            .map(|(&a, b)| a || b)
+                            .collect()
+                    }
+                    Err(error) => {
+                        warn!(
+                            "Failed pruning the partitions for expr {:?}: {}",
+                            expr, error
+                        );
+                        return partitions;
```
mildbyte, line 302: Could continue here, since the next expr might not fail?

src/provider.rs
```diff
@@ -174,6 +197,164 @@ impl TableProvider for SeafowlTable {
...
+    ) -> Result<ScalarValue> {
+        match bytes_value.as_ref() {
+            Some(bytes) => match String::from_utf8(bytes.clone()) {
+                Ok(string_val) => ScalarValue::try_from_string(string_val, data_type),
+                Err(error) => Err(DataFusionError::Internal(format!(
+                    "Failed to parse min/max value: {}",
+                    error
+                ))),
+            },
+            None => Ok(ScalarValue::Utf8(None)),
```
mildbyte, line 272: Will this None get treated as a legitimate min/max value by the pruning or will it be ignored?

migrations/postgres/20220705072248_create_tables.up.sql
```diff
@@ -43,7 +43,8 @@ CREATE TABLE physical_partition_column (
     name VARCHAR NOT NULL,
     type VARCHAR NOT NULL,
     min_value BYTEA,
-    max_value BYTEA
+    max_value BYTEA,
+    null_count INTEGER
```
mildbyte, line 47: At some point we'll have to start using actual migrations here, since currently editing the same file means we'll have to recreate existing instances from scratch. Nbd right now though.

gruuya: Yup, I realize that, thanks. I figured this would happen only after the initial release, right?

src/provider.rs
gruuya: If you're asking how does DF pruning treat None values—it treats them as unknowns, so if it's unsure that a given partition is needed it will include it (e.g. if the min max interval of a partition is `[3, None]`, it will be included for the `s2 > 5` qualifier as some rows may be required, see  [here](https://github.com/apache/arrow-datafusion/blob/3eb55e9a0510d872f6f7765b1a5f17db46486e45/datafusion/core/src/physical_optimizer/pruning.rs#L1591-L1606)).

If you're asking whether `ScalarValue::Utf8(None)` will get converted to `None` in the final min/max array of a given type, I'm unsure but plan to cover it with tests.

src/provider.rs
gruuya: We can't, because this particular filter may require a partition that no other filter does, thus skipping the partitions for this filter may lead to incorrect results. For example, for three partitions and two filters:

```python
expr1: [1, 0, 1] # first filter refutes the 2nd partition and needs the 1st and the 3rd one
expr2: [1, 1, 0] # second filter needs only the first two partitions
```
if pruning fails for the first filter and we continue, the third partition will not be included in the final partition vector.

src/provider.rs
gruuya: I tried that initially, but it was a pita to get it working; the issue is that building `ArrayRef` incrementally (like we do it in `from_partitions` by iterating over partitions/columns) is hard/unsupported, and only this conversion of the final vector via [`ScalarValue::iter_to_array`](https://github.com/apache/arrow-datafusion/blob/e0a9fa347f7adb1e6d6157b537e44c574e88244a/datafusion/common/src/scalar.rs#L806) worked.

src/provider.rs
gruuya: Indeed, the parsing was wrong; it's [fixed now](https://github.com/splitgraph/seafowl/pull/67/commits/7761b31d6d32e1c3a8b05c4f43489a69140593de), alongside with some (doc)tests that demonstrate it.

src/provider.rs
gruuya: This is doable off course, but I was thinking that going for the opposite, JIT-like, approach may be even more beneficial: don't initialize `min_values`/`max_values`/`null_counts` at all, and only upon a call to the `PruningStatistics` interface do a get-or-insert map lookup op for that particular column's requested stats. 

Since the number of columns in a table is usually much larger than the number of distinct columns used in query qualifiers I expect this to be the most performant approach.

src/lib.rs
```diff
@@ -1,3 +1,5 @@
+extern crate core;
+
```
mildbyte, line 2: Is this deliberate? I don't see `core::..` used anywhere, though googling shows people sometimes need to use it for type conversions?

src/provider.rs
gruuya: Actually, as discussed offline, you're correct; since the filters list is a big `AND` we can and should continue here (and flip the original partition mask to all 1's while using `&&` to combine the masks).

src/provider.rs
```diff
@@ -172,6 +197,199 @@ impl TableProvider for SeafowlTable {
...
+                    Ok(array) => Some(array),
+                    Err(error) => {
+                        warn!("Failed to convert vector of min/max values {:?} to array: {}", values, error);
+                        None
+                    }
+                }
+            }
+            None => None,
+        }
+    }
```
mildbyte, line 371: Dunno if this is overkill (feel free to ignore), but every time I see `None => None`, I get an urge to rewrite it

```suggestion
        values.and_then(|stats| {
            ScalarValue::iter_to_array(stats.clone().into_iter())
                .map_err(|e| {
                    warn!(
                        "Failed to convert vector of min/max values {:?} to array: {}",
                        values, e
                    );
                    e
                })
                .ok()
        })
    }
```

src/provider.rs
```diff
@@ -172,6 +197,199 @@ impl TableProvider for SeafowlTable {
...
+
+    // Prune away partitions that are refuted by the provided filter expressions
+    async fn prune(&self, filters: &[Expr]) -> Vec<SeafowlPartition> {
+        let mut partitions = self.partitions.deref().clone();
+
+        if filters.is_empty() {
+            // No qualifiers means we need all partitions
+            return partitions;
+        }
```
mildbyte, line 322: Theoretically you don't need this any more, as no filters -> partition_mask is all Trues -> will return the existing partitions. Might have a performance impact though (less so if you do iter + filter instead of clone + retain):

```suggestion
```

src/lib.rs
gruuya: Nope, either CLion or I added it accidentally.

src/provider.rs
```diff
@@ -172,6 +197,201 @@ impl TableProvider for SeafowlTable {
...
+        let mut null_counts = HashMap::new();
+
+        for field in schema.fields() {
+            min_values.insert(
+                field.name().clone(),
+                vec![ScalarValue::Utf8(None); partition_count],
+            );
+            max_values.insert(
+                field.name().clone(),
+                vec![ScalarValue::Utf8(None); partition_count],
```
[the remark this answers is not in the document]
gruuya: They should actually need to be `ScalarValue::Int16(None)`, `ScalarValue::Boolean(None)`, etc. depending on the type. But since we agreed on the JIT-like stats parsing I'll be changing that part anyway.

src/provider.rs
```diff
@@ -172,6 +197,199 @@ impl TableProvider for SeafowlTable {
...
+                        "Failed constructing a pruning predicate for expr {:?}: {}",
+                        expr, error
+                    );
+                }
+            }
+        }
+
+        let mut iter = partition_mask.iter();
+        partitions.retain(|_| *iter.next().unwrap());
+        partitions
```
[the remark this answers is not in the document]
gruuya: Right, but the docs for [`retain`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.retain) stress that it preserves the original order of elements, which made me wonder whether the other approaches don't.

src/provider.rs
gruuya: Yeah, that makes sense.

src/context.rs
```diff
@@ -158,10 +158,15 @@ fn build_partition_columns(
...
-    // TODO PhysicalPartitionColumn might not be the right data structure here (lacks ID etc)
+    // TODO PartitionColumn might not be the right data structure here (lacks ID etc)
     match &partition_stats.column_statistics {
+        // NB: Here we may end up with `null_count` being None, but DF pruning algorithm demands that
+        // the null count field be not nullable itself, and consequently for any such cases the
+        // pruning will fail, and we will default to using all partitions.
         Some(column_statistics) => zip(column_statistics, schema.fields())
             .map(|(stats, column)| {
+                // TODO: the to_string will discard the timezone for Timestamp* values, and will
+                // therefore hinder the ability to recreate them once needed for partition pruning
```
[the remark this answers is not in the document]
gruuya: Yeah, need to think about it some more.

src/provider.rs
```diff
@@ -362,4 +584,86 @@ mod tests {
...
+        "Null check zero nulls")
+    ]
+    #[test_case(
+        vec![(Some(10), Some(20), Some(0)), (Some(20), None, Some(1))],
+        vec![col("some_int").is_null()],
+        vec![1];
+        "Null check one null")
+    ]
+    // Pruning here fails with: `"Invalid argument error: Column 'some_int_null_count' is declared as non-nullable but contains null values"`
+    // so we default to using all partitions
```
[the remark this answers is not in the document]
gruuya: Even though we ignore failed expressions we still default to using all partitions in this particular case, so the comment is valid.

src/provider.rs
```diff
@@ -362,4 +584,86 @@ mod tests {
...
+        vec![col("some_int").gt_eq(lit(25))],
+        vec![1, 2];
+        "Partition with missing max")
+    ]
+    #[test_case(
+        vec![(Some(10), Some(20), None), (Some(20), Some(30), Some(0)), (Some(30), Some(40), Some(1))],
+        vec![col("some_int").gt(lit(15)), col("some_int").lt(lit(25))],
+        vec![0, 1];
+        "Multiple expressions")
+    ]
```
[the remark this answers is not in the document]
gruuya: Sure thing, adding.

tests/end_to_end.rs
```diff
@@ -313,6 +313,30 @@ async fn test_table_partitioning_and_rechunking() {
...
+
+    let expected = vec![
+        "+------------+----------------+",
+        "| some_value | some_int_value |",
+        "+------------+----------------+",
+        "| 46         | 5555           |",
+        "| 47         | 6666           |",
+        "+------------+----------------+",
+    ];
+    assert_batches_eq!(expected, &results);
```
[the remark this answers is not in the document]
gruuya: I don't think so unfortunately (`EXPLAIN ANALYZE` doesn't work).

src/context.rs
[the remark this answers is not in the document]
gruuya: Actually, since DF [doesn't even collect](https://github.com/apache/arrow-datafusion/blob/5ddad472acbbe2073aa6ac901ee5115f5830cc0f/datafusion/core/src/datasource/file_format/parquet.rs#L193-L355) statistics for timestamp types, for now this problem won't manifest itself (but we also won't be able to prune by timestamp-like columns).

mildbyte (approved):

Issue #41, opened 2022-08-03
Local object cache for S3


mildbyte:
When using an S3-compatible object store, DataFusion / Seafowl currently download the object every time they need to scan through it. We should let users dedicate some scratch space on a Seafowl instance as object cache. Potential options:

  - write a wrapper around the `ObjectStore` trait that proxies an upstream object store and uses a local FS as cache
  - use https://github.com/kahing/goofys / + catfs OR s3fs?

PR #68, opened 2022-08-24
Add support for `CREATE EXTERNAL TABLE` over HTTP

mildbyte:
- Add an HTTP/HTTPS implementation of the `ObjectStore` trait that supports Range headers 
- Wire it into DataFusion (with a small hack to get it to route all HTTP(s) URLs to our object store)
- The user can now do `CREATE EXTERNAL TABLE file STORED AS PARQUET LOCATION 'https://some-url-to.com/site/file.parquet'`

Also reorganize schemas for less confusion:

  - Remove the `datafusion` database altogether
  - Add a special `staging` schema, for and only for external tables (can't be manipulated with other DDL)
  - `CREATE EXTERNAL TABLE file` creates a table in `staging.file` (which can be referenced like that)

The tests are using wiremock and HTTP. I tested manually this works on AWS S3, though it sends a ton of small Range requests for SELECT queries, so isn't suitable for anything but ingestion.

```
    SELECT * FROM datafusion.public.supply_chains LIMIT 1 results in:
...
    bytes_scanned{filename=seafowl-public.s3.eu-west-1.amazonaws.com/tutorial/trase-supply-chains.parquet}=232699
...
    2022-08-23T19:00:58.240Z DEBUG hyper::proto::h1::io                            > flushed 171 bytes
    2022-08-23T19:00:58.283Z DEBUG hyper::proto::h1::io                            > parsed 10 headers
    2022-08-23T19:00:58.283Z DEBUG hyper::proto::h1::conn                          > incoming body is content-length (83 bytes)
    2022-08-23T19:00:58.283Z DEBUG hyper::proto::h1::conn                          > incoming body completed
    2022-08-23T19:00:58.284Z DEBUG hyper::client::pool                             > pooling idle connection for ("https", seafowl-public.s3.eu-west-1.amazonaws.com)
    2022-08-23T19:00:58.285Z DEBUG reqwest::async_impl::client                     > response '206 Partial Content' for https://seafowl-public.s3.eu-west-1.amazonaws.com/tutorial/trase-supply-chains.parquet
    2022-08-23T19:00:58.285Z DEBUG hyper::client::pool                             > reuse idle connection for ("https", seafowl-public.s3.eu-west-1.amazonaws.com)
    2022-08-23T19:00:58.286Z DEBUG hyper::proto::h1::io                            > flushed 171 bytes
    2022-08-23T19:00:58.329Z DEBUG hyper::proto::h1::io                            > parsed 10 headers
    2022-08-23T19:00:58.329Z DEBUG hyper::proto::h1::conn                          > incoming body is content-length (98 bytes)
    2022-08-23T19:00:58.329Z DEBUG hyper::proto::h1::conn                          > incoming body completed
    2022-08-23T19:00:58.329Z DEBUG hyper::client::pool                             > pooling idle connection for ("https", seafowl-public.s3.eu-west-1.amazonaws.com)
    2022-08-23T19:00:58.329Z DEBUG reqwest::async_impl::client                     > response '206 Partial Content' for https://seafowl-public.s3.eu-west-1.amazonaws.com/tutorial/trase-supply-chains.parquet
    2022-08-23T19:00:58.508Z INFO  seafowl::frontend::http                         > 127.0.0.1:59574 "POST /q HTTP/1.1" 200 "-" "curl/7.68.0" 4.063829991s
    2022-08-23T19:00:58.509Z DEBUG hyper::proto::h1::io                            > flushed 4541 bytes
    2022-08-23T19:00:58.509Z DEBUG hyper::proto::h1::conn                          > read eof
```


src/object_store/http.rs
```diff
@@ -0,0 +1,385 @@
...
+/// pass the path to us as `some-host.com/file.parquet` (as it strips the scheme and the fake host)
+/// and the object store instance will re-append the correct scheme back to the URL.
+///
+/// Returns a `None` if the location doesn't start with `http://` / `https://`.
+pub fn try_prepare_http_url(location: &str) -> Option<String> {
+    location.strip_prefix("http://").map_or_else(
+        || {
+            location
+                .strip_prefix("https://")
+                .map(|l| format!("https://{}/{}", ANYHOST, l))
```
milesrichardson, line 304: Note: An IPv6 host must be surrounded by square brackets when serialized to a URI (unless there is no port specified, in some cases?). This code looks correct, as far as I can tell.

I just felt it was worth the warning, because yesterday I spent time I will never get back on tracing this exact bug in multiple JS libraries that incorrectly serialized a URI with an IPv6 host like `::1`. (It's a new problem because Node 18 now binds to `localhost` using the resolution order of the operating system, meaning code can no longer expect `server.address()` to return an IPv4 address, even for `localhost`).

So basically I recommend adding some tests for a URI with `::1` as the host 😸

mildbyte: Thanks for the warning. We don't serialize/deserialize IPs here, we just move the existing URL around, as passed by the user. If they don't know how to serialize IPs properly, it's their fault :)

mildbyte:
Turns out, even `CREATE TABLE AS SELECT * FROM (some external table)` was being really slow, since DataFusion would run multiple small `get_range` requests for ~20KB chunks.

Wrote a cache in front of it that aligns requests to a 2MB chunk and stores them in an on-disk cache. Technically we could use it for any object store (solving https://github.com/splitgraph/seafowl/issues/41), but we won't expose settings / wrap the S3 object store with this cache in this PR.

Before, it took more than 10 minutes to do `CREATE TABLE AS` from an external table with a single 150MB Parquet file and `iftop` showed it was only using 1Mbit/s of bandwidth. Now, it takes about 40s (the bandwidth still doesn't seem to get maxed out, I'm getting about 40Mbit/s, might be due to loading happening in a single thread?)

```sql
seafowl=> SELECT volume FROM staging.supply_chains LIMIT 1;
 volume 
--------
     50
(1 row)

Time: 621.563 ms

2022-08-27T10:01:47.384Z DEBUG seafowl::object_store::cache                    > Read 1659613 bytes from the buffer for chunk 72, slicing out 1659605..1659613
 2022-08-27T10:01:47.385Z DEBUG seafowl::object_store::cache                    > Read 1659613 bytes from the buffer for chunk 72, slicing out 134634..1659605
 2022-08-27T10:01:47.404Z DEBUG seafowl::object_store::cache                    > Read 2097152 bytes from the buffer for chunk 0, slicing out 167659..192287

seafowl=> SELECT volume FROM staging.supply_chains LIMIT 1;
 volume 
--------
     50
(1 row)

Time: 70.901 ms
seafowl=> 


...

 ~ $ time curl   -H "Content-Type: application/json" localhost:8080/q   -d@- <<EOF
{"query": "CREATE TABLE supply_chains AS SELECT * FROM staging.supply_chains"}
EOF
^
real    0m37.018s
user    0m0.004s
sys     0m0.006s
```

PR #93, opened 2022-08-31
Add object garbage collection functionality

gruuya:
Add 2 new commands:
- `VACUUM PARTITIONS` to find orphan partitions (not referenced by any table version in the catalog) and delete them from the store and catalog
- `VACUUM TABLES` to delete all tables apart from the last one.

In addition, add a task that will periodically perform GC on partitions (by default the task is not run, i.e. interval = 0).


src/catalog.rs
```diff
@@ -602,6 +609,23 @@ impl PartitionCatalog for DefaultCatalog {
...
+
+    async fn delete_partitions(
+        &self,
+        object_storage_ids: Vec<String>,
+    ) -> Result<u64, Error> {
+        self.repository
+            .delete_partitions(object_storage_ids)
+            .await
+            .map_err(Self::to_sqlx_error)
+    }
```
gruuya, line 628: Perhaps instead of setting up the boilerplate catalog/interface code, it makes more sense to expose an interface for running ad-hoc queries via sqlx in the `(Default)SeafowlContext` somehow?

src/utils.rs
```diff
@@ -38,6 +40,56 @@ pub async fn run_one_off_command<W>(
     }
 }
 
+pub async fn gc_partitions(context: Arc<DefaultSeafowlContext>) {
```
gruuya, line 43: We could also run this on every `DROP TABLE` statement, though it may be an overkill. Alternatively, `sqlparser` supports the [`PURGE`](https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/ast/mod.rs#L1069) keyword for the `DROP` statement, which can be re-purposed here to initiate GC.

I also like the idea of having this be callable via a command; `sqlparser` doesn't support `VACUUM`, and the potential existing alternatives (i.e. `DEALLOCATE ALL`, `DISCARD ALL`) are not adequate I think, so we'd need to have that customized.

gruuya: As agreed, I've added the `VACUUM PARTITIONS` and `VACUUM TABLES` statement support for now.

mildbyte (commented):
Looking good! Can you also add some docs for this:

  - VACUUM syntax in https://www.splitgraph.com/docs/seafowl/reference/ddl-queries
  - new GC option in https://www.splitgraph.com/docs/seafowl/reference/seafowl-toml-configuration
  - A note in the guide (maybe https://www.splitgraph.com/docs/seafowl/guides/installing or a new "Administration/Maintenance" page) on manual/automatic garbage collection

src/utils.rs
```diff
@@ -38,6 +40,56 @@ pub async fn run_one_off_command<W>(
...
+    match context
+        .partition_catalog
+        .get_orphan_partition_store_ids()
+        .await
+    {
+        Ok(mut object_storage_ids) if !object_storage_ids.is_empty() => {
+            info!("Found {} orphan partition(s)", object_storage_ids.len());
+
+            let mut retain_map = vec![true; object_storage_ids.len()];
+            for (ind, object_storage_id) in object_storage_ids.iter().enumerate() {
```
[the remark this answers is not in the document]
gruuya: I know, I wasted some time myself and then went with this to move on to other things 😄

src/context.rs
```diff
@@ -880,7 +880,42 @@ impl SeafowlContext for DefaultSeafowlContext {
                                 output_schema: Arc::new(DFSchema::empty())
                             })),
                         }))
-                    }
+                    },
+                Statement::Truncate { table_name, partitions} => {
```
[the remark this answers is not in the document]
gruuya: Right, moving the GC execution to the physical plan part.

src/repository/default.rs
```diff
@@ -217,6 +217,23 @@ impl Repository for $repo {
         Ok((new_table_id, new_version_id))
     }
 
+    async fn delete_old_table_versions(
+        &self,
+        table_id: Option<TableId>,
+    ) -> Result<u64, Error> {
+        let query = if let Some(table_id) = table_id {
+            sqlx::query("DELETE FROM table_version WHERE id NOT IN (SELECT max(id) FROM table_version WHERE table_id = $1)").bind(table_id)
```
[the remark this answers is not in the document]
gruuya: Right, there's also a tiny problem that the present query will actually delete all versions for all tables, except the last one for the table specified 😬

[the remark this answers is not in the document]
gruuya: Ok, a more robust query pushed, as well as tests to match.

src/main.rs
```diff
@@ -187,6 +188,27 @@ Run Seafowl with --one-off instead to run a one-off command from the CLI."
...
+    // Add a GC task for purging obsolete objects from the catalog and the store
+    if config.misc.gc_interval > 0 {
+        let mut shutdown_r = shutdown.subscribe();
+        let mut interval =
+            interval(Duration::from_secs((config.misc.gc_interval * 3600) as u64));
+        tasks.push(
+            async move {
+                loop {
+                    tokio::select! {
+                        _ = interval.tick() => gc_partitions(&context).await,
```
mildbyte, line 200: Can `gc_partitions` panic for whatever reason and in that case, will this also crash the whole program? I'm not sure what the right thing to do here is. We could ignore the panic and keep going (log and GC on the next run, similarly to how the HTTP frontend does it if a request handler panics) or crash.

mildbyte (approved):

src/main.rs
gruuya: Not sure whether it can panic or not, but if it does it crashes the whole program (just tried it out).

I could wrap it up with: https://doc.rust-lang.org/std/panic/fn.catch_unwind.html?

PR #99, opened 2022-09-06
Implement multipart partition object upload

gruuya:
Instead of loading up the entire file into memory and uploading it in one big chunk, perform a multipart upload of small buffered file chunks.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -2,6 +2,7 @@
 
 use async_trait::async_trait;
 use base64::decode;
+use bytes::{BufMut, BytesMut};
 use datafusion::datasource::TableProvider;
 use datafusion::sql::ResolvedTableReference;
 use itertools::Itertools;
@@ -100,8 +101,8 @@ const MAX_ROW_GROUP_SIZE: usize = 65536;
 // - the buffer refers to the maximum size that we keep in memory at any time; there's a tradeoff at
 // play here since the smaller this is the more syscalls for reading we'll need (thus slowing us down),
 // - the chunk size denotes the size of the actual multipart request payload
-const PARTITION_OBJECT_BUFFER_SIZE: usize = 1024 * 1024;
-const PARTITION_OBJECT_CHUNK_SIZE: usize = 8 * 1024;
+const PARTITION_OBJECT_BUFFER_SIZE: usize = 128 * 1024 * 1024;
+const PARTITION_OBJECT_CHUNK_SIZE: usize = 8 * 1024 * 1024;
 
 pub fn internal_object_store_url() -> ObjectStoreUrl {
     ObjectStoreUrl::parse(format!("{}://", INTERNAL_OBJECT_STORE_SCHEME)).unwrap()
@@ -311,6 +312,8 @@ pub async fn plan_to_object_store(
     }
     writer.close().map_err(DataFusionError::from).map(|_| ())?;
 
+    warn!("Starting upload of partition objects");
+
     for partition_file_path in partition_file_paths {
         let physical = plan.clone();
         let store = store.clone();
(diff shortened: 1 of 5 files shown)


gruuya:
As of the last implementation, the memory profile for the [dataset addition](https://www.splitgraph.com/docs/seafowl/getting-started/tutorial-fly-io/part-2-deploying-to-fly-io#add-the-dataset) from the docs (`13:08:21.452Z` is the start of the actual multipart partition files upload):
image

However, when running 
```sql
CREATE EXTERNAL TABLE area1 STORED AS PARQUET LOCATION '/home/ubuntu/area1.parquet';
CREATE TABLE area1 AS SELECT * FROM staging.area1;
```
via a psql terminal (using [area1.parquet](https://rdr.ucl.ac.uk/s/5cc49e8bcc7497581b30), with size about 2.45GB) I'm seeing:
image

src/context.rs
```diff
@@ -304,6 +311,8 @@ pub async fn plan_to_object_store(
     }
     writer.close().map_err(DataFusionError::from).map(|_| ())?;
 
+    warn!("Starting upload of partition objects");
```
mildbyte, line 314:
```suggestion
    info!("Starting upload of partition objects");
```

src/context.rs
```diff
@@ -332,24 +341,57 @@ pub async fn plan_to_object_store(
...
+                                    break;
+                                } else if size != 0
+                                    && part_buffer.len() < PARTITION_FILE_MIN_PART_SIZE
+                                {
+                                    // Keep filling the chunk buffer until it surpasses the minimum required size
+                                    continue;
+                                }
+
+                                let part_size = part_buffer.len();
+                                warn!("Uploading part with {} bytes", part_size);
```
mildbyte, line 370:
```suggestion
                                info!("Uploading part with {} bytes", part_size);
```

(or even debug?)

src/lib.rs
```diff
@@ -1,3 +1,5 @@
+extern crate core;
+
```
mildbyte, line 2:
```suggestion
```

src/context.rs
```diff
@@ -332,24 +341,57 @@ pub async fn plan_to_object_store(
...
+                    let (multipart_id, mut writer) =
+                        store.inner.put_multipart(&location).await?;
+
+                    let error: std::io::Error;
+                    loop {
+                        match reader.read_buf(&mut part_buffer).await {
+                            Ok(size) => {
+                                if size == 0 && part_buffer.is_empty() {
+                                    // We've reached EOF and there are no pending writes to flush
+                                    // TODO: as per the docs size = 0 doesn't actually guarantee that we've reached EOF
```
mildbyte, line 360: Is this still true? Weird, since there's no other way to signal EOF (unless it returns a special Err type)?

gruuya:
As a comparison, the current main show the following memory profiles for the [two examples above](https://github.com/splitgraph/seafowl/pull/99#issuecomment-1239393814):
**supply_chains**
image



**area1**
image

src/context.rs
gruuya: Yes, this is still true. I added a comment with a potential workaround using [stream_len](https://doc.rust-lang.org/nightly/std/io/trait.Seek.html#method.stream_len) (`nightly-only experimental API` atm).

I've also tried to get the true file size from metadata beforehand and keep track of the total read bytes, but they never matched somehow.

gruuya:
Repeating profile once again (sanity check) for the current branch
**supply_chains**
image


**area1**
image

gruuya:
Another memory profile, with upload task semaphore, and object store multipart race fix
**suply_chains**
image


**area1**
image

PR #121, opened 2022-09-23
Add delete statement support

gruuya:
The present `DELETE` implementation is based on the fine-grained approach, whereby we utilize pruning to pinpoint which partitions will need elements removed. We then scan + filter from these partitions only (batching adjacent ones together to minimize sparse partition number) creating new substitute partitions, while re-using the ones unaffected by the `WHERE` qualifier of the `DELETE` from the old table version as is.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -48,6 +48,9 @@ use std::sync::Arc;
 
 use datafusion::common::{Column, DFField, DFSchema, ToDFSchema};
 pub use datafusion::error::{DataFusionError as Error, Result};
+use datafusion::physical_expr::create_physical_expr;
+use datafusion::physical_expr::execution_props::ExecutionProps;
+use datafusion::physical_plan::filter::FilterExec;
 use datafusion::scalar::ScalarValue;
 use datafusion::{
     arrow::{
@@ -574,7 +577,7 @@ impl DefaultSeafowlContext {
     }
 
     /// Resolve a table reference into a Seafowl table
-    fn try_get_seafowl_table(
+    pub fn try_get_seafowl_table(
         &self,
         table_name: impl Into<String> + std::fmt::Debug,
     ) -> Result<SeafowlTable> {
@@ -688,7 +691,7 @@ impl DefaultSeafowlContext {
                 // Duplicate the table version into a new one
                 new_table_version_id = self
                     .table_catalog
-                    .create_new_table_version(from_table_version)
+                    .create_new_table_version(from_table_version, true)
                     .await?;
             }
             _ => {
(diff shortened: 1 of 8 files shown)


src/context.rs
```diff
@@ -1258,17 +1285,133 @@ impl SeafowlContext for DefaultSeafowlContext {
...
+                                            &partitions.into_iter().group_by(|p| {
+                                                !partitions_to_filter
+                                                    .contains(&p.partition_id.unwrap())
+                                            })
+                                        {
+                                            partitions_grouped
+                                                .push((keep, group.collect()))
+                                        }
+
+                                        for (keep, group) in partitions_grouped {
```
mildbyte, line 1351: The code doesn't use `partitions_grouped` a second time after iterating through the `group_by` result, so you can iterate through this immediately without first persisting in `partitions_grouped`:

```suggestion
                                        let mut final_partition_ids =
                                            Vec::with_capacity(partitions.len());

                                        // Group adjacent partitions eligible for filtering
                                        for (keep, group) in
                                            &partitions.into_iter().group_by(|p| {
                                                !partitions_to_filter
                                                    .contains(&p.partition_id.unwrap())
                                            })
                                        {
```

src/context.rs
```diff
@@ -1258,17 +1285,133 @@ impl SeafowlContext for DefaultSeafowlContext {
...
+                                        warn!(
+                                                "Failed constructing pruning statistics for table {} (version: {}) during DELETE planning: {}",
+                                                table.name, table.table_version_id, error
+                                            );
+
+                                        // TODO: Fallback to scan + filter across all partitions?
+                                        return Err(DataFusionError::Execution(
+                                            "Failed executing DELETE statement"
+                                                .to_string(),
+                                        ));
```
mildbyte, line 1409: Can partition pruning fail during normal operation? I remember seeing this once:

```
 2022-09-02T15:42:54.361Z WARN  seafowl::provider       > Failed pruning the partitions for expr datepart(Utf8("YEAR"), #tpep_pickup_datetime) IN ([Int64(2020), Int64(2021)]): Error during planning: Invalid argument error: must either specify a row count or at least one column
 2022-09-02T15:42:54.381Z WARN  seafowl::provider       > Failed pruning the partitions for expr datepart(Utf8("YEAR"), #tpep_pickup_datetime) IN ([Int64(2020), Int64(2021)]): Error during planning: Invalid argument error: must either specify a row count or at least one column
 2022-09-02T15:42:54.399Z WARN  seafowl::provider       > Failed pruning the partitions for expr datepart(Utf8("YEAR"), #tpep_pickup_datetime) IN ([Int64(2020), Int64(2021)]): Error during planning: Invalid argument error: must either specify a row count or at least one column
```

If it can fail, then we should definitely treat it as if we need to inspect all partitions.

src/provider.rs
```diff
@@ -142,6 +145,48 @@ impl std::fmt::Debug for SeafowlTable {
...
+            file_groups: partitioned_file_lists,
+            statistics: Statistics::default(),
+            projection: projection.clone(),
+            limit,
+            table_partition_cols: vec![],
+        };
+
+        let format = ParquetFormat::default();
+        // TODO: filters here probably does nothing, since we handle pruning explicitly ourselves via
+        // `partitions` param
```
mildbyte, line 185: Will it prune partitions twice, doing extra work? The `filters` might be used in the `EXPLAIN` output too (though I guess it'd be inconsistent with the list of partitions, since those are already pruned)

tests/end_to_end.rs
```diff
@@ -1071,3 +1072,172 @@ async fn test_vacuum_command() {
...
+
+    assert_partition_ids(&context, 7, vec![6]).await;
+
+    // Verify results
+    let plan = context
+        .plan_query("SELECT some_value FROM test_table")
+        .await
+        .unwrap();
+    let results = context.collect(plan).await.unwrap();
+
```
mildbyte, line 1215: If `DELETE FROM ... WHERE some_value IN (43, 44)` does make partition pruning fail, this statement is a good candidate for testing the fallback case (don't use partition pruning, check all partitions).

tests/end_to_end.rs
```diff
@@ -1071,3 +1072,172 @@ async fn test_vacuum_command() {
...
+    //
+    // Execute DELETE with multiple conditions, removing entire partition 4, and trimming partitions 1 and 5
+    //
+    let plan = context
+        .plan_query("DELETE FROM test_table WHERE some_value < 43 OR some_value > 45")
+        .await
+        .unwrap();
+    context.collect(plan).await.unwrap();
+
+    assert_partition_ids(&context, 7, vec![6]).await;
```
mildbyte, line 1207: Do we end up with one partition because the pruning doesn't discard any of `1, 4, 5` and so we end up collapsing them into a single partition? It would be good to have a test where we have a `(not-pruned partition, pruned partition, not-pruned partition, not-pruned partition)` to check that we end up with `(processed partition, pruned-so-stays-the-same partition, processed-partition-last-two-combined)`

tests/end_to_end.rs
```diff
@@ -1071,3 +1072,172 @@ async fn test_vacuum_command() {
...
+    assert_partition_ids(&context, 8, vec![]).await;
+
+    // Verify results
+    let plan = context
+        .plan_query("SELECT some_value FROM test_table")
+        .await
+        .unwrap();
+    let results = context.collect(plan).await.unwrap();
+
+    assert!(results.is_empty());
```
mildbyte, line 1242: Another test suggestion: a `DELETE` that doesn't do anything (either because all partitions get pruned out or because we process partitions but nothing in there matches the qualifiers)

src/context.rs
gruuya: Actually I tried that first, but I couldn't figure out how to appease the compiler; it looks like awaits can't be used inside `group_by`'s:
```rust
error: future cannot be sent between threads safely
    --> src/context.rs:1099:41
     |
1099 |       ) -> Result<Arc<dyn ExecutionPlan>> {
     |  _________________________________________^
1100 | |         // Similarly to DataFrame::sql, run certain logical plans outside of the actual execution flow
1101 | |         // and produce a dummy physical plan instead
1102 | |         match plan {
...    |
1505 | |         }
1506 | |     }
     | |_____^ future created by async block is not `Send`
     |
     = help: within `itertools::GroupBy<bool, std::vec::IntoIter<SeafowlPartition>, [closure@src/context.rs:1339:78: 1342:46]>`, the trait `std::marker::Sync` is not implemented for `RefCell<itertools::groupbylazy::GroupInner<bool, std::vec::IntoIter<SeafowlPartition>, [closure@src/context.rs:1339:78: 1342:46]>>`
note: future is not `Send` as this value is used across an await
    --> src/context.rs:1365:50
     |
1339 | / ...                       &partitions.into_iter().group_by(|p| {
1340 | | ...                           !partitions_to_filter
1341 | | ...                               .contains(&p.partition_id.unwrap())
1342 | | ...                       })
     | |____________________________- has type `&itertools::GroupBy<bool, std::vec::IntoIter<SeafowlPartition>, [closure@src/context.rs:1339:78: 1342:46]>` which is not `Send`
...
1365 |   ...                           )
     |  ________________________________^
1366 | | ...                           .await?;
     | |____________________________________^ await occurs here, with the value maybe used later
...
1381 |   ...                   }
     |                         - the value is later dropped here
     = note: required for the cast to the object type `dyn futures::Future<Output = std::result::Result<Arc<(dyn ExecutionPlan + 'static)>, datafusion::error::DataFusionError>> + std::marker::Send`
```

src/context.rs
gruuya: The `Failed pruning the partitions for expr` signifies that the underlying DF pruning logic, called inside `SeafowlPruningStatistics`, failed for a particular expression, but in that case we just keep trying to prune partitions for other expressions (if any).

The error match above occurs when we fail to construct `SeafowlPruningStatistics` in the first place (in theory it can happen). I'll add a fallback to filtering across all partitions.

src/provider.rs
gruuya: Actually it will try to [prune row groups](https://github.com/apache/arrow-datafusion/blob/d7c0e420ce6362a253e1f55a873a0732052e0e2f/datafusion/core/src/physical_plan/file_format/parquet.rs#L424) in a single parquet file.

It is [also used](https://github.com/apache/arrow-datafusion/blob/d7c0e420ce6362a253e1f55a873a0732052e0e2f/datafusion/core/src/physical_plan/file_format/parquet.rs#L323-L331) for `EXPLAIN`.

So perhaps it is better to keep it there.

tests/end_to_end.rs
gruuya: I did add such a test; note that this is not a true no-op, as a new table version (identical to the prior one) gets created.

tests/end_to_end.rs
gruuya: That expression does not make pruning fail (as shown in the latest tests now). 

In fact, the fallback case only gets triggered when we fail to initialize `SeafowlPruningStatistics` in the first place (i.e. before we even attempt to prune). I don't see an easy way to mimic that in the tests, though I'll think some more.

src/provider.rs
```diff
@@ -101,6 +103,7 @@ impl SchemaProvider for SeafowlCollection {
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct SeafowlPartition {
+    pub partition_id: Option<PhysicalPartitionId>,
```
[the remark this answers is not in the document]
gruuya: > Is this None if we haven't saved the partition yet?

That is correct.

Honestly, I'd rather avoid re-structuring `SeafowlPartition` just for that. I'd be fine with making `partition_id` default to 0 though (and avoid using the Option/unwrap that way).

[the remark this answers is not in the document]
mildbyte: Eh, it's still better to have that explicit Option there than to overload the partition ID this way (0 might be a valid ID depending on the underlying DB). Let's keep it as an Option in that case.

tests/end_to_end.rs
mildbyte: It's also possible a newer DF version has fixed partition pruning with `IN` (the log I got is about a month old).

src/context.rs
mildbyte: Oh I see, maybe the groupby has some internal state that isn't safe to be sent across threads. You can avoid the "mut + insert" pattern by using collect() though:

```rust
diff --git a/src/context.rs b/src/context.rs
index 39d8d7e..5fc9bd6 100644
--- a/src/context.rs
+++ b/src/context.rs
@@ -1333,19 +1333,18 @@ impl SeafowlContext for DefaultSeafowlContext {
                                             );
 
                                         // Group adjacent partitions eligible for filtering
-                                        let mut partitions_grouped: Vec<(
+                                        let partitions_grouped: Vec<(
                                             bool,
                                             Vec<SeafowlPartition>,
-                                        )> = Vec::new();
-                                        for (keep, group) in
-                                            &partitions.into_iter().group_by(|p| {
+                                        )> = partitions
+                                            .into_iter()
+                                            .group_by(|p| {
                                                 !partitions_to_filter
                                                     .contains(&p.partition_id.unwrap())
                                             })
-                                        {
-                                            partitions_grouped
-                                                .push((keep, group.collect()))
-                                        }
+                                            .into_iter()
+                                            .map(|(keep, group)| (keep, group.collect()))
+                                            .collect();
 
                                         for (keep, group) in partitions_grouped {
                                             if keep {
```

mildbyte (approved):

src/context.rs
gruuya: Good point, applying patch.

Issue #104, opened 2022-09-12
Allow passing urlencoded (or otherwise escaped) values in X-Seafowl-Query


mildbyte:
Consider this SQL:

```sql
SELECT value
-- some comment
FROM some_table
```

HTTP headers don't support newlines, so our current Observable client replaces newlines with spaces. In this case, this also changes the semantic meaning of the query (the `FROM some_table` becomes a comment). This also doesn't deal with Unicode and non-printable ASCII characters. We should figure out a way to safely pass queries with any character as the header.

Headers only support printable ASCII characters. Even though it's not a direct authority on what headers proxies and `fetch()` will accept, CF's [Page Rules](https://developers.cloudflare.com/rules/transform/request-header-modification/reference/header-format/) docs are a good aproximation. So we don't need to url-encode all characters that would normally be urlencoded (e.g. spaces) to save on space.

PR #122, opened 2022-09-26
Percent-decode incoming queries

neumark:
headers cannot contain special characters such as newlines (as is described in GH issue #104). The solution is to percent-encode (aka: encodeURIComponent()) the query string. Encoding is not mandatory, some characters like the space are HTTP header-safe, these characters can be sent both as they are or percent encoded (e.g. %20 for space).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,14 +40,13 @@ datafusion = "12"
 datafusion-expr = "12"
 datafusion-proto = "12"
 futures = "0.3"
-
-# passing to DF's query planner
 hashbrown = { version = "0.12", features = ["raw"] }
 hex = ">=0.4.0"
 itertools = ">=0.10.0"
 log = "0.4"
 moka = { version = "0.9.3", default_features = false, features = ["future", "atomic64", "quanta"] }
 object_store = "0.5.0"
+percent-encoding = "2.2.0"
 pretty_env_logger = "0.4"
 prost = "0.11"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ include = [
     "Cargo.toml",
     "build.rs",
     "migrations",
+
 ]
 
 [features]
--- a/src/frontend/http.rs
+++ b/src/frontend/http.rs
@@ -19,6 +19,7 @@ use datafusion_expr::logical_plan::{LogicalPlan, PlanVisitor, TableScan};
 use futures::{future, TryStreamExt};
 use hex::encode;
 use log::{debug, info};
+use percent_encoding::percent_decode_str;
 use serde::Deserialize;
 use serde_json::json;
 use sha2::{Digest, Sha256};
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -48,6 +48,7 @@ itertools = ">=0.10.0"
 log = "0.4"
 moka = { version = "0.9.3", default_features = false, features = ["future", "atomic64", "quanta"] }
 object_store = "0.5.0"
+percent-encoding = "2.2.0"
```
mildbyte, line 51: Nice, it also looks like we're already importing this as a transitive dependency (so it's not going to result in us pulling in extra code)

```bash
~/seafowl $ cargo tree | grep percent
│   │   │   │   ├── percent-encoding v2.2.0
│   │   │   │   │   ├── percent-encoding v2.2.0
│   │   │   │   │   │   │   └── percent-encoding v2.2.0
│   │   │   │   │   │   ├── percent-encoding v2.2.0
│   │   ├── percent-encoding v2.2.0
│           └── percent-encoding v2.2.0
│   ├── percent-encoding v2.2.0
    │   │   ├── percent-encoding v2.2.0
```

src/frontend/http.rs
```diff
@@ -212,18 +213,24 @@ pub fn cached_read_query_authz(
...
     let query_hash = query_hash.split('.').next().unwrap();
 
-    let hash_str = str_to_hex_hash(&query);
+    let decoded_query = percent_decode_str(&raw_query).decode_utf8();
+
+    if decoded_query.is_err() {
+        return Err(ApiError::QueryDecodeError);
+    }
+
+    let hash_str = str_to_hex_hash(&decoded_query.as_ref().unwrap());
```
mildbyte, line 229: Instead of using `unwrap()`, you should use [the `?` operator](https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html#a-shortcut-for-propagating-errors-the--operator) to bail out with an Err value and safely extract the Ok value otherwise:

```suggestion
    let decoded_query = percent_decode_str(&raw_query).decode_utf8().map_err(|_| ApiError::QueryDecodeError)?;

    let hash_str = str_to_hex_hash(decoded_query);
```

(decoded_query should have the type String going ahead, so you don't need `unwrap()` which will panic at runtime if it encounters an Error value)

src/frontend/http.rs
```diff
@@ -562,6 +569,8 @@ mod tests {
     }
 
     const SELECT_QUERY: &str = "SELECT COUNT(*) AS c FROM test_table";
+    const PERCENT_ENCODED_SELECT_QUERY: &str = "SELECT%20COUNT(*)%20AS%20c%20FROM%20test_table";
```
mildbyte, line 572: Can you also add a case here for partial urlencoding? I'm thinking something like 

```sql
SELECT COUNT(*) AS "%F0%9F%99%82" FROM test_table
```

to test that this will still behave if we let users save space in the HTTP header by only urlencoding nonprintable/nonASCII characters

mildbyte (approved):

src/frontend/http.rs
neumark: We already kind of have that since `PERCENT_ENCODED_SELECT_QUERY` is the same as `SELECT_QUERY` with spaces replaced by `%20`. The hash of the two queries is even the same (since hash is computed after decoding).

mildbyte: Right, but I meant we should test that a space (that isn't urlencoded) still remains a space after urlencoding. But I'm now realizing that the rest of the test suite that exercises this code does that anyway (no urlencoding), so we don't need to do this.

PR #127, opened 2022-09-28
Add update statement support

gruuya:
Implementation of `UPDATE` statement in Seafowl, along similar lines as `DELETE`.

Basically, we pinpoint the exact partitions that need to be updated, and then for the new table version we do a projection of all columns that were changed using the provided assignment expressions. If a WHERE qualifier has been used, we further wrap up the assignment expressions using a CASE expression, to scope down the changes only to affected rows. The columns which were not changed are just projected as is.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -48,9 +48,8 @@ use datafusion::common::{Column, DFField, DFSchema, ToDFSchema};
 pub use datafusion::error::{DataFusionError as Error, Result};
 use datafusion::physical_expr::create_physical_expr;
 use datafusion::physical_expr::execution_props::ExecutionProps;
-use datafusion::physical_expr::expressions::{cast, Column as ColumnExpr};
+use datafusion::physical_expr::expressions::{case, cast, col};
 use datafusion::physical_plan::projection::ProjectionExec;
-use datafusion::physical_plan::union::UnionExec;
 use datafusion::physical_plan::PhysicalExpr;
 use datafusion::scalar::ScalarValue;
 use datafusion::{
@@ -1292,25 +1291,17 @@ impl SeafowlContext for DefaultSeafowlContext {
                                     partitions.iter().map(|p| p.partition_id.unwrap()),
                                 );
 
-                            // In case a WHERE qualifier has been used in the UPDATE, we'll obtain
-                            // the final output from two plans; one that will update only the rows
-                            // selected by the qualifier (using the projection_filter), and another
-                            // one that will propagate the unchanged rows (using deletion_filter).
-                            let mut projection_filter = None;
-                            let mut deletion_filter = None;
+                            let schema = table.schema().as_ref().clone();
+                            let df_schema =
+                                table.schema.arrow_schema.clone().to_dfschema()?;
+                            let mut selection_expr = None;
 
                             // Try to scope down partition ids which need to be updated with pruning
                             if let Some(expr) = selection {
-                                projection_filter = Some(create_physical_expr(
+                                selection_expr = Some(create_physical_expr(
                                     &expr.clone(),
-                                    &table.schema.arrow_schema.clone().to_dfschema()?,
-                                    table.schema().as_ref(),
-                                    &ExecutionProps::new(),
-                                )?);
-                                deletion_filter = Some(create_physical_expr(
-                                    &expr.clone().not(),
-                                    &table.schema.arrow_schema.clone().to_dfschema()?,
-                                    table.schema().as_ref(),
+                                    &df_schema,
+                                    &schema,
                                     &ExecutionProps::new(),
                                 )?);
 
@@ -1334,17 +1325,12 @@ impl SeafowlContext for DefaultSeafowlContext {
                                 }
                             }
 
-                            let schema = table.schema().as_ref().clone();
-                            let df_schema =
-                                table.schema.arrow_schema.clone().to_dfschema()?;
-
                             // Create a complete projection expression, including replicating
                             // the columns omitted in the provided assignments
                             let projection_exprs = schema
                                 .fields
                                 .iter()
-                                .enumerate()
-                                .map(|(ind, f)| {
+                                .map(|f| {
                                     if assignments.contains_key(f.name()) {
                                         let mut expr = create_physical_expr(
                                             &assignments[f.name()],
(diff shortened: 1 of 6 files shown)


src/context.rs
```diff
@@ -974,25 +997,31 @@ impl SeafowlContext for DefaultSeafowlContext {
...
                         None => None,
                         Some(expr) => Some(query_planner.sql_to_rex(expr, &table_schema, &mut HashMap::new())?),
                     };
 
-                    let assignments_expr = assignments.iter().map(|a| {
-                        Ok(Assignment { column: compound_identifier_to_column(&a.id)?, expr: query_planner.sql_to_rex(a.value.clone(), &table_schema, &mut HashMap::new())? })
-                    }).collect::<Result<Vec<Assignment>>>()?;
+                    let assignment_exprs = assignments.iter().map(|a| {
+                        Ok((
+                            table_schema.field_with_unqualified_name(&normalize_ident(&a.id[0]))?.name().clone(),
```
gruuya, line 1015: I opted for interpreting the idents as single column names, and not as compund identifiers, as it made more sense for an `UPDATE` statement.

src/context.rs
```diff
@@ -1237,52 +1266,229 @@ impl SeafowlContext for DefaultSeafowlContext {
...
+                                    }
+                                })
+                                .collect::<Result<Vec<(Arc<dyn PhysicalExpr>, String)>>>(
+                                )?;
+
+                            let mut final_partition_ids =
+                                Vec::with_capacity(partitions.len());
+
+                            // Group adjacent partitions eligible for updating
+                            let partitions_grouped: Vec<(bool, Vec<SeafowlPartition>)> =
```
gruuya, line 1383: In principle, if the ordering is not important, then we can (here as well as for DELETE) divide all the partitions into exactly two groups, and re-process the group that needs to be changed in one go.

src/context.rs
mildbyte: That makes sense, compound identifiers are only useful in an `UPDATE ... USING` (where you might update with some values derived from a different table)

tests/end_to_end.rs
```diff
@@ -1316,3 +1329,117 @@ async fn test_delete_statement() {
...
+        "|                 |                |                  |                     | 42         |",
+        "|                 | 5555           |                  | 2022-01-01 20:01:01 | 42         |",
+        "|                 | 5555           |                  | 2022-01-01 20:02:02 | 42         |",
+        "|                 | 3333           |                  | 2022-01-01 20:03:03 | 42         |",
+        "|                 | 5555           |                  |                     | 42         |",
+        "|                 | 5555           |                  |                     | 42         |",
+        "|                 |                |                  |                     | 42         |",
+        "+-----------------+----------------+------------------+---------------------+------------+",
+    ];
+    assert_batches_eq!(expected, &results);
```
mildbyte, line 1444: These tests look good. Worth testing some extra error/corner cases here:

  - requires an implicit cast (e.g. `SET some_time = '2022-01-01 21:21:21')
  - results in a type mismatch (e.g. `SET some_int_value = 'nope'`)
  - references a nonexistent column (`SET nonexistent = 42`)
  - doesn't change anything (`SET some_bool_value = TRUE WHERE some_value = 200`)
  - redundant update (`SET some_bool_value = TRUE, some_bool_value = (some_int_value = 5555)`)
  - a more complex expression (`SET some_other_value = CASE WHEN some_int_value = 5555 THEN 'five' WHEN some_int_value = 3333 THEN 'three' ELSE 'none' END`)

src/context.rs
```diff
@@ -1237,52 +1266,229 @@ impl SeafowlContext for DefaultSeafowlContext {
...
+                                        update_plan = Arc::new(ProjectionExec::try_new(
+                                            projection_exprs.clone(),
+                                            scan_plan,
+                                        )?);
+
+                                        println!("table schema: {:?}", table.schema());
+                                        println!(
+                                            "update schema: {:?}",
+                                            update_plan.schema()
+                                        );
```
mildbyte, line 1470:
```suggestion
                                        debug!("table schema: {:?}", table.schema());
                                        debug!(
                                            "update schema: {:?}",
                                            update_plan.schema()
                                        );
```

src/context.rs
mildbyte: Kinda true (especially since we're reordering data inside of the partitions anyway, since we're merging changed with the unchanged rows), but we have this code already, so I don't mind it staying here.

mildbyte (commented):
Looking good! I added suggestions for some more tests that are worth writing to expose potential issues.

src/context.rs
gruuya: Thanks, I'll rather remove this completely, this was just for debugging.

tests/end_to_end.rs
gruuya: Good idea, added all these variations now.

I've commented out the type mismatch tests for now because, even though they themselves pass, the last case errors out during planning with `Internal("Internal SQL error: \"error returned from database: syntax error at end of input\"")`. Looks like there's something not being closed/cleaned from our side while planing the type mismatch queries.

gruuya: Turns out the issue was that when planning UPDATE we first create a new table version, and only then try to execute the plan into partitions, which fails due to the type mismatch, but leaves the new empty table version hanging.

I've fixed it by creating the new table version only when needed (when linking to newly created partitions).

src/utils.rs
```diff
@@ -90,6 +92,22 @@ pub async fn gc_partitions(context: &DefaultSeafowlContext) {
     }
 }
 
+// Group adjacent partitions using the provided function
+pub fn group_partitions<F>(
+    partitions: Vec<SeafowlPartition>,
+    f: F,
+) -> Vec<(bool, Vec<SeafowlPartition>)>
+where
+    F: FnMut(&SeafowlPartition) -> bool,
```
mildbyte, line 101: Can this be `Fn` instead of `FnMut`, given that it doesn't mutate `partitions_to_filter`/`partitions_to_update`? Not sure if this will result in any compiler optimizations though.

tests/end_to_end.rs
```diff
@@ -1316,3 +1329,173 @@ async fn test_delete_statement() {
...
+
+    // This one's a bit different
+    let err = context
+        .plan_query("UPDATE test_table SET some_other_value = 'nope'")
+        .await
+        .unwrap_err();
+
+    assert!(err
+        .to_string()
+        .contains("Unsupported CAST from Utf8 to Decimal128(38, 10)"));
```
mildbyte, line 1458: Is it possible to break these out into a separate test (e.g. `test_update_errors`) just for the sake of higher test precision?

(though OTOH we do test that we can perform an update after an error, so it's worth keeping one error case here)

src/context.rs
```diff
@@ -1237,52 +1265,170 @@ impl SeafowlContext for DefaultSeafowlContext {
...
+                                    } else {
+                                        // Just replicate the omitted columns
+                                        Ok((
+                                            col(f.name(), &schema)?,
+                                            f.name().to_string(),
+                                        ))
+                                    }
+                                })
+                                .collect::<Result<Vec<(Arc<dyn PhysicalExpr>, String)>>>(
+                                )?;
```
mildbyte, line 1370: I think this `assignments -> projection exprs` function is self-contained enough that it could be factored out (and potentially unit tested) to keep this UPDATE handler shorter

mildbyte (approved):

src/utils.rs
gruuya: Yup, changing.

src/context.rs
gruuya: Good point, extracted, though note that unti testing is a bit of a pain due to need to compare trait objects for equality.

PR #128, opened 2022-09-28
Add primitive nodejs client

neumark:


clients/q.js
```diff
@@ -0,0 +1,79 @@
...
+    req.on('error', error => {
+      console.error(error);
+    });
+
+    req.end();
+};
+
+if (require.main === module) {
+    const args = process.argv.slice(2)
+    if (args[0]?.trim() === '-w') {
```
mildbyte, line 73: You can also use the POST to run a read-only SELECT query, maybe the flags should be flipped:

  - default: use POST
  - `-r` (`--read-only`): use GET

(also, the GET API isn't that useful for querying from the console if you don't have the facility to have a client side cache and perform conditional GET requests, handling ETags and 304s, but we might as well keep it here as an example).

clients/q.js
```diff
@@ -0,0 +1,79 @@
...
+
+const crypto = require('crypto');
+const http = require('http');
+
+const trimQuery = sql => sql.trim().replace(/(?:\r\n|\r|\n)/g, " ");
+
+const hash = sql => crypto.createHash('sha256').update(sql).digest('hex');
+
+const hostname = 'localhost';
+const port = 8080
```
mildbyte, line 11: We should be able to override these, e.g. to query `https://demo.seafowl.io` or a custom deployment.

mildbyte (commented):
Nice, thanks for building this! Can you put it into `examples/clients/node` instead? That's where all the example code / Compose deployments currently end up.

mildbyte (approved):

PR #135, opened 2022-10-03
Upgrade convergence to the released version

mildbyte:


mildbyte:
Looks like convergence-arrow requires a different sqlparser version (0.24) than datafusion-core (0.23), so this is currently incompatible, closing for now.

PR #147, opened 2022-10-10
Replace WASM native datatypes in UDF declarations with DDL types

neumark:


mildbyte (commented):
LGTM as discussed offline, apart from:

  - making sure old-style types (`i64`) are still accepted, since removing those will make existing instances not load at all
  - making sure this is still case-insensitive (should be, since an old test uses `I32` and `i32` interchangeably

mildbyte (approved):

PR #149, opened 2022-10-11
introduce wasi+messagepack UDF language variant

neumark:


neumark:
- [ ]  handle error messagepack-encoding arguments
- [ ]  handle WASM module loading errors
- [ ]  handle linker errors
- [ ]  handle case when UDF not found in WASM module
- [ ]  handle case when UDF invocation fails
- [ ]  handle case when messagepack decoding of output fails
- [ ]  Similar to make_scalar_function_from_wasm, this function should verify module OK and export exists up front.
- [ ]  don't panic on unexpected input / output UDF type, fail with a DataFusionError instead.
- [ ]  invoke_wasi_messagepack loads the WASM module, locates the UDF for each row, should be once per SELECT
- [ ]  Don't panic on type mismatches, because the UDF could return different types than promised

src/wasm_udf/wasm.rs
```diff
@@ -11,22 +15,518 @@ use datafusion::{
...
+        .unwrap()
+        .as_any()
+        .downcast_ref::<PrimitiveArray<T>>()
+    {
+        Some(arr) => Ok(arr.value(row_ix)),
+        None => Err(DataFusionError::Internal(format!(
+            "Error casting column {:?} to array of primitive values",
+            col_ix
         ))),
     }
```
mildbyte, line 221: I rewrote it a bit to avoid the match (mostly a stylistic choice, I didn't check if this compiles). Note that because of the `unwrap()` this function will panic if row/col_ix are out of bounds, so it's not completely "pure".

```suggestion
    let arr = args
        .get(col_ix)
        .unwrap()
        .as_any()
        .downcast_ref::<PrimitiveArray<T>>()
        .ok_or_else(|| DataFusionError::Internal(format!(
            "Error casting column {:?} to array of primitive values",
            col_ix
        )))?;
    Ok(arr.value(row_ix))
```

src/wasm_udf/wasm.rs
```diff
@@ -11,22 +15,518 @@ use datafusion::{
...
+            .unwrap()
+            .as_any()
+            .downcast_ref::<arrow::array::BooleanArray>(
+        ) {
+            Some(arr) => Ok(Value::from(arr.value(row_ix))),
+            None => Err(DataFusionError::Internal(format!(
+                "Error casting column {:?} to boolean array",
+                col_ix
+            ))),
+        },
```
mildbyte, line 278: Does the `get_arrow_value` function not work in these cases?

src/wasm_udf/wasm.rs
```diff
@@ -11,22 +15,518 @@ use datafusion::{
...
+        // deallocate both input and output buffers
+        self.dealloc
+            .call(&mut self.store, (udf_input_ptr, input_size))?;
+        self.dealloc
+            .call(&mut self.store, (udf_output_ptr, output_size))?;
+        Ok(output)
+    }
+}
+
+fn get_arrow_value<T>(
```
mildbyte, line 202: It's kind of strange that this function isn't provided by Arrow/DataFusion to begin with, it feels pretty general. Are you sure it isn't defined in a public module somewhere?

src/wasm_udf/wasm.rs
```diff
@@ -11,22 +15,518 @@ use datafusion::{
...
+    decoder: &dyn Fn(&Value) -> Result<T::Native>,
+) -> Result<PrimitiveArray<T>>
+where
+    T: ArrowPrimitiveType,
+{
+    let mut decoded_results = Vec::with_capacity(encoded_results.len());
+    for i in encoded_results {
+        decoded_results.push(Some(decoder(i)?));
+    }
+    Ok(decoded_results.iter().collect::<PrimitiveArray<T>>())
```
mildbyte, line 309: Not very keen on mut + for + push here. Also, is the `Some(...)` in the loop body doing anything? We never add a `None` to the vector anyway, so it's not like we'd be discarding items that failed to decode. Is it possible to rewrite this as:

```suggestion
    encoded_results.iter().map(decoder).collect()
```

? The `map()` will turn an iterator of `Value` into an iterator of `Result<T::Native>` and the `collect()` will turn this into a `Result<PrimitiveArray<T>>` (stopping the iteration if an error occurs, see https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect).

or, if it doesn't compile:

```suggestion
    encoded_results.iter().map(decoder).collect::<Result<PrimitiveArray<T>>>()
```

src/wasm_udf/wasm.rs
```diff
@@ -11,22 +15,518 @@ use datafusion::{
...
+            .as_i64()
+            .ok_or(DataFusionError::Internal(format!(
+                "Expected to find i64 value, but received {:?} instead",
+                v
+            ))) {
+            Err(e) => Err(e),
+            Ok(v_i64) => i16::try_from(v_i64).map_err(|e| {
+                DataFusionError::Internal(format!("Error converting i64 to i16: {:?}", e))
+            }),
+        })
```
mildbyte, line 329: You don't need this `match` here either, you can do it more cleanly as:

```suggestion
        >(encoded_results, &|v| v
            .as_i64()
            .ok_or(DataFusionError::Internal(format!(
                "Expected to find i64 value, but received {:?} instead",
                v
            ))).and_then(|v_i64| i16::try_from(v_i64).map_err(|e| {
                DataFusionError::Internal(format!("Error converting i64 to i16: {:?}", e))
            })
```

src/wasm_udf/wasm.rs
```diff
@@ -11,22 +15,518 @@ use datafusion::{
...
+                    v
+                ))) {
+                    Err(e) => Err(e),
+                    Ok(v_i64) => i32::try_from(v_i64).map_err(|e| {
+                        DataFusionError::Internal(format!(
+                            "Error converting i64 to i32: {:?}",
+                            e
+                        ))
+                    }),
+                },
```
mildbyte, line 345: something like this maybe?

```suggestion
                &|v| match v.as_i64().ok_or(DataFusionError::Internal(format!(
                    "Expected to find i64 value, but received {:?} instead",
                    v
                ))).and_then(|v_i64|  i32::try_from(v_i64).map_err(|e| {
                        DataFusionError::Internal(format!(
                            "Error converting i64 to i32: {:?}",
                            e
                        ))
                    })
```

src/wasm_udf/wasm.rs
```diff
@@ -11,22 +15,518 @@ use datafusion::{
...
+
+fn messagepack_decode_results(
+    return_type: &CreateFunctionDataType,
+    encoded_results: &Vec<Value>,
+) -> Result<ArrayRef> {
+    match return_type {
+        CreateFunctionDataType::SMALLINT => decode_udf_result_primitive_array::<
+            arrow::datatypes::Int16Type,
+        >(encoded_results, &|v| match v
+            .as_i64()
```
mildbyte, line 320: Does messagepack not let you do `as_i16`/`as_i32` directly?

src/wasm_udf/wasm.rs
```diff
@@ -11,22 +15,518 @@ use datafusion::{
...
+            for i in encoded_results {
+                decoded_results.push(Some(i.as_str().ok_or(
+                    DataFusionError::Internal(format!(
+                        "Expected to find string value, received {:?} instead",
+                        &i
+                    )),
+                )?));
+            }
+            Ok(Arc::new(decoded_results.iter().collect::<StringArray>()) as ArrayRef)
+        }
```
mildbyte, line 374: Does `decode_udf_result_primitive_array` not work here because it's not a `PrimitiveArray`? You could also rewrite this function to avoid mut + for:

```suggestion
            let decoded_results = encoded_results.iter().map(|i| (i.as_str().ok_or(
                DataFusionError::Internal(format!(
                    "Expected to find string value, received {:?} instead",
                    &i
                )),
            ))).collect::<Result<StringArray>>()?;
            Ok(Arc::new(decoded_results) as ArrayRef)
        }
```

src/wasm_udf/wasm.rs
```diff
@@ -11,22 +15,518 @@ use datafusion::{
...
+                        "Expected to find bool value, received {:?} instead",
+                        &i
+                    )),
+                )?));
+            }
+            Ok(Arc::new(
+                decoded_results
+                    .iter()
+                    .collect::<arrow::array::BooleanArray>(),
+            ) as ArrayRef)
```
mildbyte, line 415: similar to above (assuming that `Some(...)` wrapper is redundant):

```suggestion
            let decoded_results = encoded_results.iter().map(|i| (i.as_bool().ok_or(
                DataFusionError::Internal(format!(
                   "Expected to find bool value, received {:?} instead",
                    &i
                )),
            ))).collect::<Result<BooleanArray>>()?;
            Ok(Arc::new(decoded_results) as ArrayRef)
```

src/wasm_udf/wasm.rs
```diff
@@ -11,22 +15,518 @@ use datafusion::{
...
+            .map(|a| Arc::new(a) as ArrayRef)
+        }
+        CreateFunctionDataType::F32
+        | CreateFunctionDataType::REAL
+        | CreateFunctionDataType::FLOAT => decode_udf_result_primitive_array::<
+            arrow::datatypes::Float32Type,
+        >(encoded_results, &|v| match v {
+            Value::F32(n) => Ok(*n),
+            _ => Err(DataFusionError::Internal(format!(
+                "Expected to find f64 value, but received {:?} instead",
```
mildbyte, line 437:
```suggestion
                "Expected to find f32 value, but received {:?} instead",
```

src/wasm_udf/wasm.rs
```diff
@@ -11,22 +15,518 @@ use datafusion::{
...
+                    "Expected to find str value, received {:?} instead",
+                    &i
+                )))?;
+                decoded_results.push(Some(s.parse::<i128>().map_err(|e| {
+                    DataFusionError::Internal(format!(
+                        "Error parsing string to i128: {:?}",
+                        e
+                    ))
+                })?));
+            }
```
mildbyte, line 458: This is definitely untested :joy: I'm happy with your variant if it works. Though I'm not sure about how this decimal type even works. Do the `precision`/`scale` values matter here and should the guest function know them in order to be able to decode the i128 into an actual "decimal" value it can do work on?

```suggestion
            let decoded_results = encoded_results.iter().map(|i|
                i.as_str().ok_or(DataFusionError::Internal(format!(
                    "Expected to find str value, received {:?} instead",
                    &i
                ))).and_then(|s| s.parse::<i128>().map_err(|e|
                    DataFusionError::Internal(format!(
                        "Error parsing string to i128: {:?}",
                        e
                )))))
            .collect::<Result<arrow::array::Decimal128Array>>()?;
            Ok(Arc::new(decoded_results) as ArrayRef)


```

src/wasm_udf/wasm.rs
```diff
@@ -11,22 +15,518 @@ use datafusion::{
...
+    // returning a Result.
+    let function_name = function_name.to_owned();
+    let module_bytes = module_bytes.to_owned();
+    let _outer_instance = WasmMessagePackUDFInstance::new(&module_bytes, &function_name)
+        .map_err(|err| {
+            DataFusionError::Internal(format!(
+                "Error initializing WASM + MessagePack UDF {:?}: {:?}",
+                function_name, err
+            ))
+        })?;
```
mildbyte, line 485: This variable seems to be unused?

src/wasm_udf/wasm.rs
```diff
@@ -94,52 +594,36 @@ fn make_scalar_function_from_wasm(
         let mut results: Vec<Val> = Vec::new();
         results.resize(array_len, Val::null());
 
-        for i in 0..array_len {
+        for row_ix in 0..array_len {
             let mut params: Vec<Val> = Vec::with_capacity(args.len());
             // Build a slice of WASM Val values to pass to the function
-            for j in 0..args.len() {
```
mildbyte, line 100: Nice refactor!

src/wasm_udf/wasm.rs
```diff
@@ -432,4 +944,132 @@ c40201087f230041206b2203240020032002370318200320013703102003\
...
+            FROM text_values;",
+            )
+            .await
+            .unwrap()
+            .collect()
+            .await
+            .unwrap();
+
+        let expected = vec![
+            "+-----+-------+---------------+",
```
mildbyte, line 1064: Sweet to see it work end to end!

src/wasm_udf/wasm.rs
```diff
@@ -11,22 +15,518 @@ use datafusion::{
...
+            Ok(DataType::Timestamp(TimeUnit::Nanosecond, None))
+        }
+    }
+}
+
+fn get_wasm_module_exported_fn<Params, Results>(
+    instance: &Instance,
+    store: &mut Store<WasiCtx>,
+    export_name: &str,
+) -> anyhow::Result<TypedFunc<Params, Results>>
```
mildbyte, line 68: Is the `anyhow` crate necessary here at all? In other places, we either return a `DataFusionError` directly or define a specific error type, e.g. https://github.com/splitgraph/seafowl/blob/main/src/frontend/http_utils.rs#L49-L116

src/wasm_udf/wasm.rs
```diff
@@ -11,22 +15,518 @@ use datafusion::{
...
+        // TODO: handle linker errors
+        let instance = linker.instantiate(&mut store, &module)?;
+
+        let alloc = get_wasm_module_exported_fn(&instance, &mut store, "alloc")?;
+        let dealloc = get_wasm_module_exported_fn(&instance, &mut store, "dealloc")?;
+        let udf = get_wasm_module_exported_fn(&instance, &mut store, function_name)?;
+        let memory = match instance.get_memory(&mut store, "memory") {
+            Some(mem) => Ok(mem),
+            None => Err(anyhow::anyhow!("could not find module's exportd memory")),
+        }?;
```
mildbyte, line 114: Some more code golf (though I think this can be a DataFusionError instead):

```suggestion
        let memory = instance.get_memory(&mut store, "memory").ok_or(||
            anyhow::anyhow!("could not find module's exported memory")
        )?;
```

mildbyte (commented):
I left a few suggestions on how to potentially clean some things up and remove some redundant code. Since we only test a few primitive types, I've no idea if most of these serialize/deserialize correctly. It would be nice to have a bigger test that uses a function that tries to work on weirder data types (like date, timestamp, boolean, decimal).

src/wasm_udf/wasm.rs
```diff
@@ -11,22 +15,528 @@ use datafusion::{
...
+        let ptr: usize = udf_output_ptr.try_into().unwrap();
+        const SIZE_BYTE_COUNT: usize = std::mem::size_of::<i32>();
+        let mut size_buffer = [0u8; SIZE_BYTE_COUNT];
+        self.memory
+            .read(&self.store, ptr, &mut size_buffer)
+            .map_err(|err| {
+                anyhow::anyhow!(format!("error reading output buf size: {:?}", err))
+            })?;
+        let size: usize = i32::from_ne_bytes(size_buffer).try_into().unwrap();
+        //eprintln!("read_udf_output(): output buffers size: {:?} bytes", size);
```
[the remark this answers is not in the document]
mildbyte: (I wrote the review before these eprintln!s were gone, but you could still bring them back as debug statements if needed)

src/wasm_udf/wasm.rs
```diff
@@ -11,22 +15,528 @@ use datafusion::{
...
+        let (output, output_size) = self.read_udf_output(udf_output_ptr)?;
+        //eprintln!("call(): read UDF output {:?} bytes", output_size);
+        // deallocate both input and output buffers
+        //eprintln!("call(): deallocating udf input and output buffers");
+        self.dealloc
+            .call(&mut self.store, (udf_input_ptr, input_size))?;
+        self.dealloc
+            .call(&mut self.store, (udf_output_ptr, output_size))?;
+        Ok(output)
+    }
```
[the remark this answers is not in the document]
mildbyte: (I wrote the review before these eprintln!s were gone, but you could still bring them back as debug statements if needed)

src/wasm_udf/wasm.rs
neumark: That's the part where the module's exports are verified before constructing `inner` to fail early if there's something wrong with the wasm module (eg: no export named `concat2`).

src/wasm_udf/wasm.rs
neumark: thx!

src/wasm_udf/wasm.rs
neumark: Good point, got rid of `anyhow`

src/wasm_udf/wasm.rs
neumark: Totally agree; I don't know enought about Arrow/DataFusion to be sure, but I couldn't find it anywhere.

src/wasm_udf/wasm.rs
neumark: I struggled with this for a while, the resulting code is definitely a compromise:
- `decoder()` can return an error, in which case there's no point in continuing a loop. I don't know of a way to terminate `iter().collect()` early.
- One option is to collect a default value for the type, but I couldn't get the Rust compiler to accept my attempt to get this working (and in retrospect it seems cleaner to exist in case of error instead of packing a vector full of defaults).
- The `Some()` is actually required by the call to `collect()`, not entirely sure why tbh.

src/wasm_udf/wasm.rs
neumark: Definitely looks nicer, thanks!

src/wasm_udf/wasm.rs
neumark: No, not sure why. I think the reason might be that it has variable length integer encodings and the serialized size of the integer (which is input dependent) shouldn't leak into the types used in the code (which is checked compile-time). Instead, it's all hid behind `i64`.

src/wasm_udf/wasm.rs
neumark: Oooh, nice catch! 😅

src/wasm_udf/wasm.rs
neumark: I'm not entirely sure. It's telling that's `Decimal128Type::Native = i128` and there's no clear logic involving precision and scale in that file. My guess is that it's only display related (but I could be wrong ofc).

mildbyte: Looks like the precision/scale do have semantic meaning:

https://arrow.apache.org/docs/python/generated/pyarrow.decimal128.html

The scale is basically by how many digits the decimal is shifted:

> As an example, decimal128(7, 3) can exactly represent the numbers 1234.567 and -1234.567 (encoded internally as the 128-bit integers 1234567 and -1234567, respectively), but neither 12345.67 nor 123.4567.
> decimal128(5, -3) can exactly represent the number 12345000 (encoded internally as the 128-bit integer 12345), but neither 123450000 nor 1234500.

src/wasm_udf/wasm.rs
neumark: Looks a lot nicer this way!

src/wasm_udf/wasm.rs
mildbyte: I see, looks like it's dual with encoding back from messagepack to arrow (where StringArray/BooleanArray aren't primitive arrays)

src/wasm_udf/wasm.rs
mildbyte: (the Some is required as we discovered but it can be re-added inside of the closure)

mildbyte (approved):
LGTM, great work! This will need some more testing (especially with decimals which I'm now convinced won't work in UDFs without the UDF somehow knowing the precision), but we can figure this out later.

neumark:
> 

Agreed

PR #154, opened 2022-10-17
Table version querying aka time travel

gruuya:
This PR presents the initial implementation of Seafowl table time travel functionality, allowing one to reference previous (non-latest, which is the default) table versions using table functions SQL syntax: `SELECT * FROM my_table(version)`.

A couple of notes on how this works:
- A new mutable query visitor has been implemented, that walks through all nodes in the sqlparser's `Statement::Query` AST tree corresponding to `SELECT`'s. Note that I've purposefully pruned the visitor so that only the nodes under the `Query` enum variant are visited for now, as those are only needed for time travel (in order to reach any table identifer references). In theory, since many of the leaf nodes don't contain table references the visitor code could be pruned even further.
- When doing logical planning, we do an initial walk on the `SELECT` AST, recording any tables version references in the statement; if none are found, we continue processing as before (i.e. ask the DF planner to generate a plan for the statement). If we do find table version refs, we then:
- Load all table version ids (and the corresponding creation timestamps) from the DB for the tables we detected have version refs in the query.
- Parse the versions found in the query. At the moment we only support timestamp string in one particular format (RFC 3339), but that can be easily extended to a lot of other timestamp formats (using chrono's `parse_from_str`), interval specifiers (e.g. `-2d` for two days ago). Potentially, very useful could be a relative notation `SELECT * FROM my_table(-2)` (the function arg is parsed without error as `UnaryOp { op: Minus, expr: Value(Number("2", false)) }`), whereby we would select the version 2 versions prior to the latest one (if it exists), and interpret the positive values as starting from the oldest available version.
- When the version is parsed, we triage the exact table version id that corresponds to it. For timestamps this simply means the latest version at that point in time.
- Once we have collected all referenced table version ids, we do another pass over the AST, this time renaming the versioned tables to `table:<version_id>`.
- Finally, we load the tables from the DB and register them into the present context with the above naming convention (the benefit here is that we have only one addition for many potential version id references that point to the same `version_id`), thus updating the schema provider's tables map. DF planner takes it from there, and any queries will be using the schema/partitions for the particular table version specified.


src/repository/default.rs
```diff
@@ -79,12 +80,57 @@ impl Repository for $repo {
     async fn get_all_columns_in_database(
         &self,
         database_id: DatabaseId,
+        table_version_ids: Vec<TableVersionId>,
```
mildbyte, line 83: Minor: use `Option<Vec<TableVersionId>>` and `None` to denote "get all versions"? It's kinda weird to use the empty list to denote all version IDs (though one could argue using `None` for that is as weird)

src/repository/interface.rs
```diff
@@ -295,7 +310,7 @@ pub mod tests {
         // Test loading all columns
```
mildbyte, line 310: Add a test here for a specific table version to make sure the query is correct for PG and SQLite?

src/version.rs
```diff
@@ -0,0 +1,263 @@
...
+            rewrite_ready: false,
+        }
+    }
+
+    pub fn table_with_version(&self, name: &ObjectName, version: &str) -> String {
+        let version_id = self.table_versions[&(name.clone(), version.to_owned())];
+        format!("{}:{}", name.0.last().unwrap().value, version_id.unwrap())
+    }
+
+    // // Try to parse the specified version timestamp into a Unix epoch
```
mildbyte, line 47:
```suggestion
    // Try to parse the specified version timestamp into a Unix epoch
```

src/version.rs
```diff
@@ -0,0 +1,263 @@
...
+    ]
+    #[test_case(
+        "SELECT * FROM some_schema.test_table";
+        "Basic select with schema and table name")
+    ]
+    #[test_case(
+        "SELECT * FROM some_db.some_schema.test_table";
+        "Basic select with fully qualified table name")
+    ]
+    fn test_table_version_rewrite(query: &str) {
```
mildbyte, line 229: I'd add a test case here for something like a CTE or a table name that doesn't reference a version at all, something like:

```sql
WITH some_cte AS (SELECT 1 AS k) SELECT t.*, k.* FROM some_schema.test_table('2022-01-01 ...') t JOIN some_cte c ON t.k = c.k
```

src/version.rs
```diff
@@ -0,0 +1,263 @@
...
+        );
+        // Do a first pass
+        rewriter.visit_query(&mut q);
+
+        // Set a mock table version id and do another pass for actually renaming the table
+        let id = 42 as TableVersionId;
+        for table_version_id in rewriter.table_versions.values_mut() {
+            *table_version_id = Some(id);
+        }
+        rewriter.rewrite_ready = true;
```
mildbyte, line 257: Just to confirm my understanding, during normal operation this is done by `triage_version_ids()`?

tests/end_to_end.rs
```diff
@@ -1616,3 +1620,120 @@ async fn test_update_statement_errors() {
...
+        &version_timestamps,
+    )
+    .await;
+    query_table_version(
+        &context,
+        4 as TableVersionId,
+        &version_results,
+        &version_timestamps,
+    )
+    .await;
```
mildbyte, line 1738: Can you also check `information_schema.tables/columns` here to see if the versioned tables pollute the advertised list of tables in the database?

src/repository/default.rs
gruuya: Fine with me, changing.

src/version.rs
gruuya: That is correct.

tests/end_to_end.rs
gruuya: Yup, it's not polluted; added a test for it.

src/version.rs
gruuya: Added tests for this now.

tests/end_to_end.rs
mildbyte: Ah, makes sense since we reset the context before every query. Can something like this happen though? Assume we're running this in a single thread and, because of async code, we end up interleaving as follows:

- Query 1 (Q1) comes in: `SELECT * FROM some_table('2020-01-01...')`
- Query 2 (Q2) comes in: `SELECT * FROM information_schema.tables`
- Database struct prepared for Q2, context points to it: C->D2 (some_table)
- Database struct prepared for Q1, context points to it: C->D1 (some_table)
- Processing for Q1 to insert table versions: C -> D1(some_table, some_table:42)
- Q2 executes, lists all tables, since C is pointing to D1 we get `some_table, some_table:42`

gruuya: Good point, I guess that could happen. Even worse, if a query comes in after `Processing for Q1 to insert table versions: C -> D1(some_table, some_table:42)`, but before executing Q1, it will reset the map (removing `some_table:42`), and Q1 will error out during execution due to the missing table ref.

gruuya: An easy potential remedy for this (I think) would be to simply clone the inner context when we encounter a version reference, create a new query planner from that cloned inner context's state, and register the table into the cloned context:
```diff
diff --git a/src/context.rs b/src/context.rs
index 26e0ece..1fcec94 100644
--- a/src/context.rs
+++ b/src/context.rs
@@ -855,6 +855,9 @@ impl SeafowlContext for DefaultSeafowlContext {
                     version_processor.visit_query(&mut q);

                     if !version_processor.table_versions.is_empty() {
+                        let inner = self.inner.clone();
+                        let state = inner.state.read().clone();
+
                         version_processor.triage_version_ids(self.table_catalog.clone()).await?;
                         // We now have table_version_ids for each table with version specified; do another
                         // run over the query AST to rewrite the table.
@@ -871,9 +874,12 @@ impl SeafowlContext for DefaultSeafowlContext {
                             let table_ref = TableReference::from(full_name.as_str());
                             let table_provider = tables_by_version[&table_version_id.unwrap()].clone();

-                            if !self.inner.table_exist(table_ref)? {
-                                self.inner.register_table(table_ref, table_provider)?;
+                            if !inner.table_exist(table_ref)? {
+                                inner.register_table(table_ref, table_provider)?;
                             }
+
+                            let query_planner = SqlToRel::new(&state);
+                            return query_planner.sql_statement_to_plan(Statement::Query(q));
                         }
                     }
```

mildbyte: Makes sense, let's go for that just to be safe.

mildbyte (approved):

PR #168, opened 2022-10-24
System schema and table version table

gruuya:
Implement a new systems schema, allowing for adding custom view-like tables that will query our metadata store and provide relevant insight to the users. The first example introduced here is `table_versions` table, which can be used to browse the available table versions and their creation timestamps.

### DIFF
--- a/src/system_tables.rs
+++ b/src/system_tables.rs
@@ -1,4 +1,4 @@
-//! Implementation for creating virtual Seafowl system tables, inspired by influxdb_iox system tables
+//! Mechanism for creating virtual Seafowl system tables, inspired by influxdb_iox system tables
 //! and datafusion's information_schema.
 
 use crate::catalog::TableCatalog;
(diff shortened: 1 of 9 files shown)


tests/end_to_end.rs
```diff
@@ -214,15 +215,19 @@ async fn test_create_table() {
     let results = list_columns_query(&context).await;
```
gruuya, line 215: Perhaps better to omit the system schema from this and `list_tables_query`, and rely on explicit tests instead?

src/system_tables.rs
```diff
@@ -0,0 +1,187 @@
...
+            // database_name field, since we scope down to the database at hand.
+            database,
+            schema: Arc::new(Schema::new(vec![
+                Field::new("table_schema", DataType::Utf8, false),
+                Field::new("table_name", DataType::Utf8, false),
+                Field::new("table_version_id", DataType::Int64, false),
+                Field::new(
+                    "creation_time",
+                    // TODO: should we be using a concrete timezone here?
+                    DataType::Timestamp(TimeUnit::Second, None),
```
gruuya, line 134: I can't see a Timestamp builder that uses a timezone so I went with None atm.

tests/end_to_end.rs
```diff
@@ -1832,29 +1861,56 @@ async fn test_table_time_travel() {
...
+    let expected = vec![
+        "+--------------+------------+------------------+",
+        "| table_schema | table_name | table_version_id |",
+        "+--------------+------------+------------------+",
+        "| public       | test_table | 1                |",
+        "| public       | test_table | 2                |",
+        "| public       | test_table | 3                |",
+        "| public       | test_table | 4                |",
+        "| public       | test_table | 5                |",
+        "+--------------+------------+------------------+",
```
mildbyte, line 1913: Another test here that filters on a version ID / creation timestamp, just to make sure the rest of DataFusion projection / filtering machinery works as expected (which it should, since `TableVersionsTable` implements the table trait)?

tests/end_to_end.rs
mildbyte: Maybe, but other tests that don't care about `information_schema` (e.g. `test_create_external_table_http`) have it in there as well, so let's leave it like that.

mildbyte (approved):

PR #226, opened 2022-11-28
Remote table WHERE clause pushdown

### LINKED ISSUE #224: Filter pushdown in remote tables
Investigate options and implement pushdown of the `WHERE` clause during remote table scans. The solution should ensure that:
1. Given an Expr, we can easily determine whether it is push-able to a remote or not. We could use DataFusion's `ExpressionVisitor` for that, or do something more explicit (along the lines of Debug implementation for Expr). This could be used in `supports_filter_pushdown` to make DF plan the scan accordingly.
2. Closely related, and crucially we should be able to reconstruct the SQL from the given Expr. Perhaps we can have something like
```rust
pub trait SqlRepresentation: fmt::Display {
    fn to_sql(&self) -> String {
        format!("{}", self)
    }
}
```
   
and then override it for DF structs that don't have adequate representation. In addition, it seems natural to have this done at the some time while checking for push-ability. In case we'd want to do it all through `ExpressionVisitor` we'd need some kind of template engine to collect the final SQL string since the recursion there is not explicit as for Display/Debug.

3. Make the default mechanism overridable for particular remote database system, in case we want e.g. a different operator symbol from the default one.

gruuya:
Implement a FDW-like mechanism to enable pushdown of filter qualifiers to a remote data source as an optimization to skip redundant row fetching over the network.

The mechanism is based on the DataFusion `ExpressionVisitor`, which allows us to walk the filter expression AST, and determine if all entities are ship-able, and if so convert them if needed to the source native form. The conversion and the entire ship-ability logic is customizable.

Closes #224.


src/remote_tables/provider.rs
```diff
@@ -175,4 +214,17 @@ impl TableProvider for RemoteTable {
...
+        &self,
+        filter: &Expr,
+    ) -> Result<TableProviderFilterPushDown> {
+        // TODO: add caching of filter to expr mapping (here and in scan)?
+        if self.filter_expr_to_sql(filter).is_none() {
+            return Ok(TableProviderFilterPushDown::Unsupported);
+        }
+        // TODO: make this Inexact just to be on the safe side? The downside is that DF won't pushdown
+        // any LIMIT clause to the `scan` then, since it and WHERE are not commutable
+        Ok(TableProviderFilterPushDown::Exact)
```
mildbyte, line 228: If we can only convert some expressions to the remote table's language, will DataFusion just try to push just those down? In other words, if some expressions are un-push-downable, does that ruin the entire pushdown?

Re: exact/inexact, I'd err on the side of exact and making `filter_expr_to_sql` return None if we're not sure that the expression will filter everything cleanly. We could also make `filter_expr_to_sql` itself return Exact/Inexact/Unsupported.

Re: caching, it's probably fine, we do this planning once per query, so I don't really anticipate it being the bottleneck.

src/remote_tables/pushdown_visitor.rs
```diff
@@ -0,0 +1,204 @@
...
+            Operator::BitwiseXor => Some("^".to_string()),
+            _ => Some(op.to_string()),
+        }
+    }
+}
+
+pub trait FilterPushdownVisitor {
+    fn scalar_value_to_sql(&self, value: &ScalarValue) -> Option<String> {
+        match value {
+            ScalarValue::Utf8(Some(val)) => Some(format!("'{}'", val)),
```
mildbyte, line 51: I think the single quotes should be escaped: `val.replace("'", "''")` (python-like syntax)

src/remote_tables/pushdown_visitor.rs
```diff
@@ -0,0 +1,204 @@
...
+            _ => Some(format!("{}", value)),
```
mildbyte, line 52: I think some non-UTF scalar values still need to be in single quotes, e.g. datetimes in PostgreSQL. More here: https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-CONSTANTS

Perhaps we could make this safer by always emitting a cast (e.g. `'2022-01-01 12:34:56'::timestamp`) but then we'd need to know the database-specific type information.

src/remote_tables/pushdown_visitor.rs
```diff
@@ -0,0 +1,204 @@
...
+                self.pushdown_supported = false;
+                return Ok(Recursion::Stop(self));
+            }
+        };
+        Ok(Recursion::Continue(self))
+    }
+
+    fn post_visit(mut self, expr: &Expr) -> Result<Self> {
+        match expr {
+            Expr::Column(col) => self.sql_exprs.push(col.name.clone()),
```
mildbyte, line 83: We should consider quoting column names here (in PG, the identifiers are quoted with double quotes, e.g. `'"' + col.replace('"', '""') + '"'"`); in MySQL they use backticks (`)

src/remote_tables/pushdown_visitor.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+}
+
+// Walk the filter expression AST for a particular remote source type and see if the expression is
+// ship-able, at the same time converting elements (e.g. operators) to the native representation if
+// needed.
+pub fn filter_expr_to_sql<T: FilterPushdownVisitor>(
+    filter: &Expr,
+    source_pushdown: T,
+) -> Result<String> {
```
mildbyte, line 142: Currently it looks like an `Err(...Execution)` is going to be silently swallowed by `filter_map`, so we can't distinguish between pushdowns being unsupported for a certain operator and genuine errors (if those can be raised by this function). Should this be an Option instead, if this function can't error out?

src/remote_tables/pushdown_visitor.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    #[rstest]
+    #[case::simple_binary_expression(
+        col("a").gt_eq(lit(25)),
+        "a >= 25")
+    ]
+    #[case::complex_binary_expression(
+        or(and(or(col("a").eq(lit(1)), col("b").gt(lit(10))), col("c").lt_eq(lit(15.0))), col("d").not_eq(lit("some_string"))),
+        "(a = 1 OR b > 10) AND c <= 15 OR d != 'some_string'")
+    ]
```
mildbyte, line 188: Can you also add a couple of tests here for:

  - strings with single quotes in them
  - timestamps
  - column names with spaces getting quoted properly

(they won't be integration tests but we'll be able to have some cases for them being escaped / serialized correctly)

src/remote_tables/pushdown_visitor.rs
```diff
@@ -0,0 +1,204 @@
...
+            Expr::Literal(val) => {
+                let sql_val = self.source.scalar_value_to_sql(val).ok_or_else(|| {
+                    DataFusionError::Execution(format!(
+                        "Couldn't convert ScalarValue {:?} to a compatible one for the remote system",
+                        val,
+                    ))
+                })?;
+                self.sql_exprs.push(sql_val)
+            }
+            Expr::BinaryExpr(be @ BinaryExpr { .. }) => {
```
mildbyte, line 93: This is out of scope for this PR, but I think eventually we'll want to support function pushdown. Something like `date_trunc` / `date_part` is very useful in reporting queries (e.g. I can get data for a certain month) but won't be pushed down right now.

src/remote_tables/pushdown_visitor.rs
```diff
@@ -0,0 +1,204 @@
...
+                    .pop()
+                    .unwrap_or_else(|| panic!("Missing left sub-expression of {}", expr));
+
+                // Similar as in Display impl for BinaryExpr: since the Expr has an implicit operator
+                // precedence we need to convert it to an explicit one using extra parenthesis if the
+                // left/right expression is also a BinaryExpr of lower operator precedence.
+                if let Expr::BinaryExpr(right_be @ BinaryExpr { .. }) = &*be.right {
+                    let p = right_be.precedence();
+                    if p == 0 || p < be.precedence() {
+                        right_sql = format!("({})", right_sql)
```
mildbyte, line 110: You could also always use brackets for order of operations and not bother with precedence (and potential bugs), since nobody will probably be looking at the actual generated SQL (we're also assuming here that DataFusion's operator precedence rules are the same as the remote database).

mildbyte (approved):

mildbyte:
Thanks! :)

src/remote_tables/pushdown_visitor.rs
mildbyte: Follow-up: maybe there's some way we could use an ORM/SQL query builder in `sqlx` for a particular dialect to get around some database-level syntax differences.

src/remote_tables/pushdown_visitor.rs
gruuya: Yup, there's a bunch of other expressions that are not ship-able atm, such as `IsNull`, `InList`, `Exists` and other `Expr` enum variants. The plan is to iteratively add support for those eligible in subsequent PRs.

src/remote_tables/provider.rs
gruuya: Yes, DataFusion optimizer is [smart enough](https://github.com/apache/arrow-datafusion/blob/master/datafusion/optimizer/src/filter_push_down.rs#L736-L756) to pushdown as much as possible and keep what it needs to ensure result correctness. It seems to split the filters up by the `AND`'s, since then it can still retain benefit from partial filter pushdown. 

Here's an example; for the following query
```sql
SELECT "date field", c FROM staging.remote_table WHERE (a > 2 OR c = 'two') AND (b < 5.0) LIMIT 2
```
the calls to `supports_filter_pushdown` and `scan` are as follows
```bash
supports_filter_pushdown is Exact: staging.remote_table.a > Int32(2) OR staging.remote_table.c = Utf8("two")
supports_filter_pushdown is Unsupported: CAST(staging.remote_table.b AS Float64) < Float64(5)
supports_filter_pushdown is Exact: staging.remote_table.a > Int32(2) OR staging.remote_table.c = Utf8("two")
supports_filter_pushdown is Exact: staging.remote_table.a > Int32(2) OR staging.remote_table.c = Utf8("two")
supports_filter_pushdown is Unsupported: CAST(staging.remote_table.b AS Float64) < Float64(5)
supports_filter_pushdown is Exact: staging.remote_table.a > Int32(2) OR staging.remote_table.c = Utf8("two")
scan received filter: a > Int32(2) OR c = Utf8("two")
final query: SELECT "b", "c", "date field" FROM "source table" WHERE a > 2 OR c = 'two'
```
The second filter is not shipable since there is no match for the CASE node in the filter expression visitor.

On the other hand, for
```sql
SELECT "date field", c FROM staging.remote_table WHERE (a > 2 OR c = 'two') AND (a < 5) LIMIT 2
```
I see
```bash
supports_filter_pushdown is Exact: staging.remote_table.a > Int32(2) OR staging.remote_table.c = Utf8("two")
supports_filter_pushdown is Exact: staging.remote_table.a < Int32(5)
supports_filter_pushdown is Exact: staging.remote_table.a > Int32(2) OR staging.remote_table.c = Utf8("two")
supports_filter_pushdown is Exact: staging.remote_table.a < Int32(5)
supports_filter_pushdown is Exact: staging.remote_table.a > Int32(2) OR staging.remote_table.c = Utf8("two")
supports_filter_pushdown is Exact: staging.remote_table.a < Int32(5)
scan received filter: a > Int32(2) OR c = Utf8("two")
scan received filter: a < Int32(5)
final query: SELECT "c", "date field" FROM "source table" WHERE a > 2 OR c = 'two' AND a < 5 LIMIT 2
```
You can see that not only have both filters been pushed down, the `LIMIT` clause was too.

As for caching, the number of visitor invocations per query per filter is 3-5, but the AST of the filter is bound to be very shallow so for now I think it's ok not to cache it. We can easily change that later on.

src/remote_tables/pushdown_visitor.rs
gruuya: I prefer keeping it `Result<String>` so as to propagate.expose that and any visitor errors. I've added a change where this is used for logging.

src/remote_tables/pushdown_visitor.rs
gruuya: Source-compatible identifier quotation and escaping added now.

src/remote_tables/pushdown_visitor.rs
mildbyte: Makes sense, though note that this still might introduce bugs due to operator precedence mismatch (but low risk)

src/remote_tables/provider.rs
```diff
@@ -87,21 +89,22 @@ impl RemoteTable {
...
-            SourceType::Postgres => {
-                let postgres_pushdown = PostgresFilterPushdown {};
-                filter_expr_to_sql(filter, postgres_pushdown).ok()
+        let result = match self.source_conn.ty {
+            SourceType::Postgres => filter_expr_to_sql(filter, PostgresFilterPushdown {}),
+            SourceType::SQLite => filter_expr_to_sql(filter, SQLiteFilterPushdown {}),
+            SourceType::MySQL => filter_expr_to_sql(filter, MySQLFilterPushdown {}),
+            _ => {
+                debug!(
+                    "Filter not shipable due to unsupported source type {:?}",
```
mildbyte, line 98:
```suggestion
                    "Filter not shippable due to unsupported source type {:?}",
```

src/remote_tables/pushdown_visitor.rs
mildbyte: Oh yeah, no need to overcomplicate this PR by adding an ORM.

mildbyte (approved):

src/remote_tables/pushdown_visitor.rs
gruuya: Tests added now.

PR #231, opened 2022-12-02
Extend filter expression pushdown surface area

gruuya:
Add pushdown of boolean filter expressions as well as the in-list expression.

### DIFF
--- a/src/remote_tables/pushdown_visitor.rs
+++ b/src/remote_tables/pushdown_visitor.rs
@@ -13,6 +13,16 @@ pub struct FilterPushdown<T: FilterPushdownVisitor> {
     pub sql_exprs: Vec<String>,
 }
 
+impl<T: FilterPushdownVisitor> FilterPushdown<T> {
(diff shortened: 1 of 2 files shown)


src/remote_tables/pushdown_visitor.rs
```diff
@@ -13,6 +14,16 @@ pub struct FilterPushdown<T: FilterPushdownVisitor> {
     pub sql_exprs: Vec<String>,
 }
 
+impl<T: FilterPushdownVisitor> FilterPushdown<T> {
+    // Intended to be used in the node post-visit phase, ensuring that SQL representation of inner
+    // nodes is on the stack.
+    fn pop_sql_expr(&mut self) -> String {
+        self.sql_exprs
+            .pop()
+            .unwrap_or_else(|| panic!("No SQL expression in the stack"))
```
mildbyte, line 23: I think this could work too:

```suggestion
            .expect("No SQL expression in the stack")
```

mildbyte (approved):

Issue #175, opened 2022-10-26
Investigate support for Delta Tables as backing storage


rupurt:
Howdy,

Are there any plans to support Delta Tables? This could work really well with GraphQL subscriptions.

mildbyte:
Hey! Do you mean being able to support DataBricks' Delta Tables / Delta Lake (https://github.com/delta-io/delta/blob/master/PROTOCOL.md) as a storage backend / data source for `CREATE EXTERNAL` tables?

Design-wise, a GraphQL frontend is a sweet idea, though I'm not sure how to make it work well for analytical/aggregation queries (e.g. being able to represent a `group by` or `window` on arbitrary columns as a set of supported GraphQL fields). Same with subscriptions -- how would you quickly update a result for `AVG(volume) GROUP BY country_id`? IIRC ClickHouse/Materialize did some heavy research in that direction -- would indeed be cool to have it also available to Web devs via GQL :)

gruuya:
@rupurt thanks for the very cool ideas! :)

As for using Delta tables for our storage backend/layer (i.e. replacing our [DIY lakehouse protocol](https://www.splitgraph.com/blog/seafowl-time-travel) with the Delta one using delta-rs), this is something that we'll likely converge towards at some point later on.

For now though, with the latest Seafowl version (`0.2.10`) you should be able to [instantiate the delta tables](https://seafowl.io/docs/reference/ddl-queries#delta-tables) stored in various cloud object stores as an external table (will be placed in the `staging` schema) and query them.

rupurt:
Amazing. Thank you @gruuya

gruuya:
> As for using Delta tables for our storage backend/layer (i.e. replacing our [DIY lakehouse protocol](https://www.splitgraph.com/blog/seafowl-time-travel) with the Delta one using delta-rs), this is something that we'll likely converge towards at some point later on.

I'm happy to say that we've [completed](https://github.com/splitgraph/seafowl/issues/324) this migration, so this issue can be closed now. Thanks for a great idea @rupurt !

PR #252, opened 2022-12-27
Enable external Delta tables in Seafowl

gruuya:
Utilize the Delta table factory for DataFusion, and the new options mechanism for creating external Delta tables in Seafowl. 

Example definition for AWS S3-backed delta table:
```sql
CREATE EXTERNAL TABLE my_delta
STORED AS DELTATABLE
OPTIONS ('AWS_ACCESS_KEY_ID' 'secret', 'AWS_SECRET_ACCESS_KEY' 'also_secret', 'AWS_REGION' 'eu-west-3') 
LOCATION 's3://my-bucket/my-delta-table/'
```


mildbyte (approved):
Looks great! Do you know if the new deps make the final binary size much larger? Might be worth putting the support behind the Cargo features mechanism so that people can turn it off if necessary.

mildbyte:
(tagging https://github.com/splitgraph/seafowl/issues/175 since this partially solves it)

gruuya:
Good point, thanks!

I've featurized remote/delta table querying mechanisms now (both are default), and measured the binary sizes in a couple of scenarios:

1. all features—65MB
2. all features except delta tables—58MB
3. all features except delta and remote tables—55MB

So delta tables contribute about 10% to the new binary size, with remote tables being much more lightweight.

PR #269, opened 2023-01-18
Minimal fix for unquote bug for DataFusion 15.0

neumark:
There's a [nasty bug](https://github.com/apache/arrow-datafusion/blob/3cc607de4ce6e9e1fd537091e471858c62f58653/datafusion/sql/src/statement.rs#L167) in `datafusion-sql` which results in quoted collection and column names being parsed incorrectly (with the `"` character being considered part of the name).

During Splitgraph to Seafowl sync, the following SQL statements are run:
```sql
CREATE EXTERNAL TABLE tmp_hvlfgdshkjhnkbhy STORED AS PARQUET LOCATION 'https://storage.googleapis.com/jruw973mdb/p.parquet';
CREATE TABLE tmp_hvlfgdshkjhnkbhy AS SELECT * FROM staging.tmp_hvlfgdshkjhnkbhy;
CREATE SCHEMA "neumarktest/seafowlsync";
ALTER TABLE tmp_hvlfgdshkjhnkbhy RENAME TO "neumarktest/seafowlsync"."table1";
```

This MR requires to absolute minimum required to enable the query above to run.
Independetly, I will submit a fix to the datafusion table, hopefully included in DF 16+, which will fix the quoting issues when we upgrade rendering this code obsolete.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -35,7 +35,8 @@ use mockall::automock;
 use object_store::{path::Path, ObjectStore};
 
 use sqlparser::ast::{
-    AlterTableOperation, ObjectType, Statement, TableFactor, TableWithJoins,
+    AlterTableOperation, Ident, ObjectName, ObjectType, SchemaName, Statement,
+    TableFactor, TableWithJoins,
 };
 
 use arrow_integration_test::field_to_json;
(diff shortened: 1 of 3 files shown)


src/context.rs
```diff
@@ -127,6 +128,39 @@ fn quote_ident(val: &str) -> String {
...
+            SchemaName::Simple(remove_quotes_from_object_name(schema_name))
+        }
+        SchemaName::UnnamedAuthorization(auth) => {
+            SchemaName::UnnamedAuthorization(remove_quotes_from_ident(auth))
+        }
+        SchemaName::NamedAuthorization(schema_name, auth) => {
+            SchemaName::NamedAuthorization(
+                remove_quotes_from_object_name(schema_name),
+                remove_quotes_from_ident(auth),
+            )
```
mildbyte, line 159: Are these ones necessary / used by us?

mildbyte (approved):

src/context.rs
neumark: Not that I know of. I thought de-quoting them might be useful to avoid a future footgun, but since this code is hopefully a temporary fix until DF 16, I agree it's better to remove it.

gruuya (approved):
Nice catch!

As mentioned offline, it would be great to also cover the `CREATE EXTERNAL TABLE` path with this fix, and also alter one of the integration tests to exercise it.

PR #270, opened 2023-01-20
Upgrade to DF 16+ version

gruuya:
Upgrade Seafowl to Datafusion post 16.0.0 that includes the fix needed to convert statements to plan from the session state.


src/context.rs
```diff
@@ -565,11 +560,18 @@ impl DefaultSeafowlContext {
...
+        &self,
+        name: &OwnedTableReference,
+    ) -> Result<OwnedTableReference> {
+        // NB: Since Datafusion 16.0.0 there's this OwnedTableReference enum and for external tables
+        // the parsed ObjectName (which may be multipart, fully-qualified name) is coerced into the
+        // `Bare` enum variant (see `external_table_to_plan` in datafusion-sql) for some reason.
+        //
+        // This means that any potential catalog/schema references get condensed into the name, so
+        // we have to unravel that name here again, and then resolve it properly.
+        let full_name = name.to_string();
```
neumark, line 573: Due to a datafusion bug (still unfixed in 16.0), calling `ObjectName.to_string()` will double quote names with special chars. Instead, we must do:
```rust
pub fn remove_quotes_from_ident(possibly_quoted_name: &Ident) -> Ident {
    Ident::new(&possibly_quoted_name.value)
}

pub fn remove_quotes_from_idents(column_names: &[Ident]) -> Vec<Ident> {
    column_names.iter().map(remove_quotes_from_ident).collect()
}

pub fn remove_quotes_from_object_name(name: &ObjectName) -> ObjectName {
    ObjectName(remove_quotes_from_idents(&name.0))
}

// ...
let full_name = remove_quotes_from_object_name(&name).to_string()
```

These fuctions are added to `context.rs` in [PR 269](https://github.com/splitgraph/seafowl/pull/269).

src/context.rs
```diff
@@ -565,11 +560,18 @@ impl DefaultSeafowlContext {
...
+        let reference = TableReference::from(full_name.as_str());
```
neumark, line 574: Not sure if this as an issue: If I have a table called `my.table`, then could this incorrectly be parsed as `my` being the collection / schema? Disregard if "." is not a valid char even within quotes for a table name.

src/context.rs
```diff
@@ -914,11 +952,16 @@ impl SeafowlContext for DefaultSeafowlContext {
         statement: DFStatement,
     ) -> Result<LogicalPlan> {
         // Reload the schema before planning a query
+        // TODO: A couple of possible optimisations here:
+        // 1. Do a visit of the statement AST, and then load the metadata for only the referenced identifiers.
+        // 2. No need to load metadata for the TableProvider implementation maps when instantiating SqlToRel,
+        //    since it's sufficient to have metadata for TableSource implementation in the logical query
+        //    planning phase. We could use a liter structure for that, and implement `ContextProvider` for
```
neumark, line 959: Apologies for bringing up spelling: "liter" -> "lighter"?

src/wasm_udf/data_types.rs
```diff
@@ -62,28 +62,22 @@ pub enum CreateFunctionDataType {
...
-    fn default() -> Self {
-        CreateFunctionVolatility::Volatile
-    }
-}
 
 #[derive(Serialize, Deserialize, Debug, PartialEq, Eq, EnumString, Display, Clone)]
 #[serde(rename_all = "camelCase")]
+#[derive(Default)]
 pub enum CreateFunctionLanguage {
+    #[default]
```
neumark, line 77: I realize this is just `clippy` rewriting the existing code, but can we make `WasmMessagePack` the default?

neumark (approved):
Very impressive achievement, the upgrade impact is enormous!
Congrats!!!

src/context.rs
gruuya: Understood, though note that the `name` above is actually not an `ObjectName`, but an `OwnedTableReference` instead, and it's [`to_string()`](https://github.com/apache/arrow-datafusion/blob/fad77a4b278a441e343fb6caa76f6e3bcd27e248/datafusion/common/src/table_reference.rs#L107-L121) shouldn't require unquoting right?

Regardless, I'll wait for your PR to merge before merging this one.

src/context.rs
gruuya: Good point. I just tried it out and indeed we there is this problem. Namely if I do
```sql
CREATE EXTERNAL TABLE \"my.table\"
...
```
The parsed reference has `default` for the catalog, but `"my` as the schema and `table"` as the table name.

src/context.rs
gruuya: No worries, fixing 👍🏻

src/object_store/cache.rs
```diff
@@ -419,8 +419,8 @@ mod tests {
     #[case::fetch_nothing_at_the_chunk_boundary(16..16)]
     #[case::fetch_one_byte_at_the_chunk_boundary(16..17)]
     #[case::part_of_chunk_0_part_of_chunk_1(10..20)]
-    #[case::reaches_the_end_of_the_body(10..500)]
-    #[case::goes_beyond_the_end_of_the_body(400..505)]
+    #[case::reaches_the_end_of_the_body(10..484)]
```
mildbyte, line 422: Wait, what happened here? Did something change in the `object_store`?

mildbyte (approved):

src/object_store/cache.rs
gruuya: Good question, I'm not quite sure. My initial impression is that something changed in the (`ArrowWriter` of the) `parquet` crate, but didn't dig into it further.

mildbyte: Ah, makes sense. I forgot this was using a Parquet file for tests.

PR #289, opened 2023-02-07
Implement multi-DB support in Seafowl

### LINKED ISSUE #288: Add support for connecting to other databases
The default GET & POST endpoints only support querying the default DB.

gruuya:
Specifically, add support to scope down the session context to a particular database for a given statement/batch of statements. In other words cross-DB queries are not supported atm.  This is done by extending the existing GET & POST routes to also support the prefixed `/db-name/q[/query-hash]` form, and then effectively switch the default database to the one provided.

Also contains the implementation of `CREATE DATABASE` statement.

The upload endpoint needs to be altered as well, but I plan on doing it in a separate PR.

Closes #288.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -116,3 +116,4 @@ vergen = "7"
 [profile.release]
 codegen-units = 1
 lto = true
+strip = true
--- a/src/context.rs
+++ b/src/context.rs
@@ -8,7 +8,7 @@ use datafusion::datasource::TableProvider;
 use datafusion::parquet::basic::Compression;
 use itertools::Itertools;
 use object_store::local::LocalFileSystem;
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use tokio::fs::File as AsyncFile;
 use tokio::io::{AsyncReadExt, AsyncWriteExt, BufReader};
 
@@ -85,6 +85,7 @@ use datafusion_expr::logical_plan::{
 };
 use datafusion_expr::{DmlStatement, Filter, WriteOp};
 use log::{debug, info, warn};
+use parking_lot::RwLock;
 use prost::Message;
 use tempfile::TempPath;
 use tokio::sync::Semaphore;
@@ -256,6 +257,7 @@ pub struct DefaultSeafowlContext {
     pub internal_object_store: Arc<InternalObjectStore>,
     pub database: String,
     pub database_id: DatabaseId,
+    pub all_database_ids: Arc<RwLock<HashMap<String, DatabaseId>>>,
     pub max_partition_size: u32,
 }
 
@@ -511,6 +513,16 @@ pub fn is_statement_read_only(statement: &DFStatement) -> bool {
     }
 }
 
+// The only reason to keep this trait around (instead of migrating all the functions directly into
+// DefaultDeafowlContext), is that `create_physical_plan` would then be a recursive async function,
+// which works for traits, but not for structs: https://stackoverflow.com/a/74737853
+//
+// The workaround would be to box a future as the return of such functions, which isn't very
+// appealing atm (involves heap allocations, and is not very readable).
+//
+// Alternatively, if we know that all recursive calls could be handled by the inner (DataFusion's)
+// `create_physical_plan` we could also rewrite the calls explicitly like that, and thus break the
+// recursion.
 #[cfg_attr(test, automock)]
 #[async_trait]
 pub trait SeafowlContext: Send + Sync {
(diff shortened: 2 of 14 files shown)


Cargo.toml
```diff
@@ -116,3 +116,4 @@ vergen = "7"
 [profile.release]
 codegen-units = 1
 lto = true
+strip = true
```
mildbyte, line 119: Oh, nice, looks like the `true` value [strips all symbols](https://doc.rust-lang.org/cargo/reference/profiles.html#strip) as well.

src/context.rs
```diff
@@ -556,17 +568,55 @@ pub trait SeafowlContext: Send + Sync {
...
+            .inner()
+            .copied_config()
+            .with_default_catalog_and_schema(name.clone(), DEFAULT_SCHEMA);
+
+        Ok(Arc::from(DefaultSeafowlContext {
+            inner: SessionContext::with_config_rt(
+                session_config,
+                self.inner().runtime_env(),
+            ),
+            table_catalog: self.table_catalog.clone(),
```
mildbyte, line 595: Does this mean we load the _entire_ world (the state of all tables and all schemas in all databases) and then scope the context down to a single DB? Would it be possible to load just the context for a single database at the beginning of handling the HTTP request (we know its name already and we wouldn't support cross-db queries).

src/frontend/http.rs
```diff
@@ -420,10 +440,24 @@ pub fn filters(
...
 
+    // TODO: The path parameter parsing below is very unergonomic, mostly because warp doesn't
+    // provide better primitives (e.g. `warp::generic::Either` is private outside of the crate).
+    // The logic is to try and match the endpoint with or without the optional prefix for the DB
+    // name, and then map the first parameter in both cases to the same `Option<String>` type, which
+    // lets us use the `unify` filter method.
+
     // Cached read query
     let ctx = context.clone();
-    let cached_read_query_route = warp::path!("q" / String)
```
mildbyte, line 425: Maybe you could use something like `warp::path!("q" / String)[....].then(uncached_read_write_query)`, `warp::path!("q" / String / String)[...].then(uncached_read_write_query_db)` and the two functions would be thin wrappers that call the actual function with a `Some(dbname)`/`None` respectively? Not sure if that would be cleaner though, since you'd be defining four routes.

src/frontend/http.rs
```diff
@@ -108,12 +110,21 @@ async fn physical_plan_to_json(
...
-/// POST /q
+/// POST /q or /[database_name]/q
 pub async fn uncached_read_write_query(
+    database_name: Option<String>,
     user_context: UserContext,
     query: String,
-    context: Arc<dyn SeafowlContext>,
+    mut context: Arc<DefaultSeafowlContext>,
 ) -> Result<Vec<u8>, ApiError> {
+    // If a specific DB name was used as a parameter in the route, scope the context to it,
```
mildbyte, line 120: We might need to create the DB lazily here on first write, also because if we need to use `/q` to create the DB first and there are a lot of databases, the context will have to load all of their tables/schemas.

src/frontend/http.rs
```diff
@@ -629,29 +724,45 @@ mod tests {
...
     #[tokio::test]
-    async fn test_get_cached_write_query_error() {
-        let context = in_memory_context_with_single_table().await;
+    async fn test_get_cached_write_query_error(
+        #[values(None, Some("test_db"))] new_db: Option<&str>,
+    ) {
+        let context = in_memory_context_with_single_table(new_db).await;
         let handler = filters(context, http_config_from_access_policy(free_for_all()));
 
+        let mut uri = format!("/q/{}", str_to_hex_hash(CREATE_QUERY));
```
mildbyte, line 735: You could factor these out into a tiny helper function `fn make_uri(query: &str, db: Option<&str>) -> String` to avoid some repetition.

src/frontend/http.rs
gruuya: Yeah, I thought about this and it seemed like a lot more verbose solution.

I'm ok with keeping it as is for now, hopefully simplifying it one day to something like
```rust
let cached_read_query_route = warp::path!("q" / String)
        .or(warp::path!(String / "q" / String))
        .and(warp::get())
        .map(|either| match either {
            Either::A(query_hash) => (None, query_hash),
            Either::B((db_name, query_hash)) => (Some(db_name), query_hash),
        })
        ...
        .then(cached_read_query)
        .map(into_response);
```
conditional on `warp::generic::Either` becoming public (I think it's private only due to an oversight).

src/context.rs
gruuya: Actually no, we don't load the entire world. First of, reminder that `{table, partition, function}_catalog` is only a [thin client](https://github.com/splitgraph/seafowl/blob/631de991a9dc84b7497a041c4423795f0d1d12c1/src/config/context.rs#L34-L72) for fetching the corresponding metadata during planning/execution, the only thing it stores inside is a repository (with a connection pool). These are all shared via the Arc.

The actual state is kept in the inner `SessionContext` (and the `database`/`database_id`). When we perform scoping we start off from a [blank state](https://github.com/splitgraph/seafowl/blob/multi-db-impl/src/context.rs#L591-L594) (containing only the new default DB name and default schema). The state of all existing schemas/tables is then loaded as before prior to planning each query via [reload_schema](https://github.com/splitgraph/seafowl/blob/multi-db-impl/src/context.rs#L1037). Thus the only "penalty" for targeting a different DB is re-creating a new blank `SessionContext`, no actual additional DB hits required since `CREATE DATABASE` ensures we have the name/id entry in the shared `all_database_ids` map.

So we do load the context for a single database only, but we do it per-statement not per-HTTP request (this was also the case before). This means we'll perform multiple `reload_schema`'s while processing consecutive statements from a statement batch, which I think is fine for now (and can be changed later on).

src/frontend/http.rs
gruuya: As mentioned context loads schemas/tables for a single DB only, and this is the case also for the default route `/q`, i.e. we load `DEFAULT_DB` stuff in the `reload_schema`. Also, note that `CREATE DATABASE` works in any reference system, so I don't think we need to create the DB here.

src/context.rs
mildbyte: Ah, sweet! Thanks for clarifying.

src/frontend/http.rs
mildbyte: Got it -- so for the replication use case would we be able to somehow determine if a DB exists and then create it if not, e.g. by trying to first `POST` a `SELECT 1` to `/q/db-id` or by hitting the information schema?

gruuya: Yeah, `POST` of `SELECT 1` to `/db-id/q` would work for that. Alternatively, catch `Unknown database my-db; try creating one with CREATE DATABASE first` error during initial replication attempt, create the DB and then retry replication.

Another option would be to have a separate endpoint to list all available DBs.

mildbyte (approved):

Issue #57, opened 2022-08-17
More sophisticated ETag calculation


mildbyte:
Follow-up to https://github.com/splitgraph/seafowl/issues/20

Currently, we compute the ETag based on all versions of Seafowl tables in a query. This disregards:

  - Contents changing when the version doesn't (e.g. using https://www.splitgraph.com/docs/seafowl/guides/baking-dataset-docker-image will always use V1 for all tables, since we're rebuilding the dataset from scratch each time): we should use the hashes of the Parquet files instead?
  - UDF/built-in function definitions changing
  - DataFusion query plan / execution changing

rupurt:
A prolly tree data structure could come in useful here https://github.com/attic-labs/noms/blob/master/doc/intro.md#prolly-tree-structure.

If a query plan can be evaluated deterministically it could use chunk hashing during each stage of applying the operations to pull results from an internal cache.

Are you familiar with doltdb https://github.com/dolthub/dolt? Essentially I would love to apply those semantics to seafowl!!

ignoramous:
Another method: A weak etag with timestamp baked in can then let clients decide whether the information they might fetch from caches is within an acceptable threshold or not. If not, they issue a request that bypasses caches.

PR #307, opened 2023-03-03
Migration of storage layer to delta-rs

gruuya:
This is the initial effort of switching our custom MVCC/Lakehouse protocol to use the Delta lake protocol instead.

A couple of general points:
- Full range of DDL & DML statements are supported, including `DELETE` & `UPDATE`
- `VACUUM` still not implemented for new tables
- Likewise, GC of dropped new tables is not implemented yet, though the needed foundation is there (`dropped_table` table and related methods)
- No perf testing has been done yet (the object upload code is likely sub-optimal as it doesn't do [multipart](https://github.com/splitgraph/seafowl/pull/99) upload)
- cleanup of legacy code hasn't been completely done yet: I only replaced the stuff that was in the way of getting things working with delta-rs, in theory the entire writing logic for Seafowl tables can be erased now. The reading logic for Seafowl tables should still remain (there are tests for that too), as the plan is to have this available throughout the transient Seafowl versions for easier data migration.
- I've also bumped to Datafusion 19.0.0/arrow post-33.0.0 in order to pick up https://github.com/apache/arrow-rs/pull/3752 and https://github.com/delta-io/delta-rs/pull/1212.

As for the implementation details:
- I've opted for keeping a single root object store around (the `inner` in `internal_object_store`), and then "scoping" it for each DeltaTable accordingly. This means wrapping it up with some other object stores so that `/` points to the table root uri (see below). The other alternative that delta-rs offers is providing the store_options, out of which it will then build the store for a given uri, but I didn't like this mostly because the memory store implementation then doesn't work (it get's re-created from scratch for each create/write/table builder instantiation).
- The table root is set to `<inner_store_root>/<table_uuid>/` (that is the place where delta-rs puts all the table-related stuff as per the protocol). For instance `s3://my_bucket/e62c26e2-b6e8-414a-a37a-66dd77430b5d/` or `file:///data/dir/e62c26e2-b6e8-414a-a37a-66dd77430b5d/`. The `table_uuid` is kept in the `table` catalog table and created randomly for each new table.
- This approach means that dropping a schema/table can be done lazily (e.g. during subsequent `VACUUM`s in chunked batches) because table uri is unique to a given table and independent of db/schema/table name. Likewise, renaming a table is only a catalog operation.


mildbyte (commented):
This looks like it's on the right track, great work!

Re: the `/[dbname]/[schemaname]/[tablename]` vs `/[table_uuid]` path prefix schema, I'm inclined towards the latter. It looks like it'd help us avoid a lot of complexity and edge cases with managing object store prefixes and paths. Re: potential downsides:

- We'll have to have the catalog DB anyway to store UDF definitions and future metadata (unless we somehow manage to keep everything in the delta lake as flat files and avoid the catalog DB altogether, which could be an interesting thought experiment)
- The delta spec doesn't seem to concern itself with how the tables are named/arranged (it only specifies what happens inside the table directory), so I don't think many other tools would rely on a specific way the table directory is named/accessed vs. just operating on the contents of that directory
- Depending on the object store implementation, there might be some argument against having a giant root directory with all the tables vs a hierarchy of `db/schema/table` (performance reasons, fewer items in a single directory)
- Similarly, we might be able to use some object store specific authorization code / IAMs to implement multitenancy -- this might mean sharding just by db-id/table-id? Not sure.

How do information schema queries work under this proposed model? It looks like we use the catalog DB to satisfy them instead of crawling the object store for all tables and their schemas?

src/provider.rs
```diff
@@ -91,7 +92,9 @@ impl CatalogProvider for SeafowlDatabase {
 
 pub struct SeafowlCollection {
     pub name: Arc<str>,
-    pub tables: RwLock<HashMap<Arc<str>, Arc<SeafowlTable>>>,
+    // TODO: consider using DashMap instead of RwLock<HashMap<_>>: https://github.com/xacrimon/conc-map-bench
+    pub legacy_tables: RwLock<HashMap<Arc<str>, Arc<SeafowlTable>>>,
+    pub tables: RwLock<HashMap<Arc<str>, Arc<dyn TableProvider>>>,
```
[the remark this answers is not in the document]
gruuya: Probably; I also don't remember why `Arc<str>`.

gruuya:
> How do information schema queries work under this proposed model? It looks like we use the catalog DB to satisfy them instead of crawling the object store for all tables and their schemas?

Yes, that is correct, we pre-load the schema information from our catalog.

src/system_tables.rs
```diff
@@ -282,3 +309,85 @@ impl SeafowlSystemTable for TablePartitionsTable {
...
+            // database_name field, since we scope down to the database at hand.
+            database,
+            schema: Arc::new(Schema::new(vec![
+                Field::new("table_schema", DataType::Utf8, false),
+                Field::new("table_name", DataType::Utf8, false),
+                Field::new("uuid", DataType::Utf8, false),
+                Field::new("deletion_status", DataType::Utf8, false),
+                Field::new(
+                    "drop_time",
+                    // TODO: should we be using a concrete timezone here?
```
mildbyte, line 333: 🤷 could split the difference and return the server's local timezone (if SQLite even supports storing timestamps like that) or following what the rest of Delta does

src/repository/default.rs
```diff
@@ -595,12 +644,86 @@ impl Repository for $repo {
...
+        &self,
+        maybe_table_id: Option<TableId>,
+        maybe_collection_id: Option<CollectionId>,
+        maybe_database_id: Option<DatabaseId>,
+    ) -> Result<(), Error> {
+        // Currently we hard delete only legacy tables, the others are soft-deleted by moving
+        // them to a special table that is used for lazy cleanup of files via `VACUUM`.
+        // TODO: We could do this via a trigger, but then we'd lose the ability to actually
+        // perform hard deletes at the DB-level.
+        // NB: We rally only need the uuid for cleanup, but we also persist db/col name on the off
```
mildbyte, line 666:
```suggestion
        // NB: We really only need the uuid for cleanup, but we also persist db/col name on the off
```

src/catalog.rs
```diff
@@ -358,6 +385,30 @@ impl DefaultCatalog {
...
+    ) -> (Arc<str>, Arc<dyn TableProvider>)
+    where
+        I: Iterator<Item = &'a AllDatabaseColumnsResult>,
+    {
+        let table_columns_vec = table_columns.collect_vec();
+
+        // Build a delta table but don't load it yet; we'll do that only for tables that are
+        // actually referenced in a statement, via the async `table` method of the schema provider.
+
+        // TODO: if the table has no columns, the result set will be empty, so we use the default UUID (all zeros).
```
mildbyte, line 401: Are there any implications? do we use this UUID to apply some DDL modifications to 0-column tables later down the line?

src/context.rs
```diff
@@ -514,14 +530,239 @@ pub fn is_statement_read_only(statement: &DFStatement) -> bool {
     }
 }
 
+// Appropriated from https://github.com/delta-io/delta-rs/pull/1176; once the DELETE and UPDATE ops
```
mildbyte, line 533: I'd move these to something like `/src/delta-rs/backports.rs` like we have with some of datafusion's parser code to make it explicit this is temporary code and to avoid ballooning this file's line length.

src/context.rs
```diff
@@ -741,10 +1033,69 @@ impl DefaultSeafowlContext {
...
+
+        let delta_schema = DeltaSchema::try_from(schema)?;
+
+        // TODO: we could be doing this inside the DB itself (i.e. `... DEFAULT gen_random_uuid()`
+        // in Postgres and `... DEFAULT (uuid())` in SQLite) however we won't be able to do it until
+        // sqlx 0.7 is released (which has libsqlite3-sys > 0.25, with the SQLite version that has
+        // the `uuid()` function).
+        // Then we could create the table in our catalog first and try to create the delta table itself
+        // with the returned uuid (and delete the catalog entry if the object store creation fails).
+        // On the other hand that would complicate etag testing logic.
```
mildbyte, line 1045: It's kind of related to this: https://github.com/splitgraph/seafowl/issues/57 -- the etags should be based on the content hash, rather than on the table version ID, in which case we wouldn't need the UUID to be deterministic.

src/context.rs
```diff
@@ -741,10 +1033,69 @@ impl DefaultSeafowlContext {
...
+        #[cfg(test)]
+        let table_uuid = deterministic_uuid();
```
mildbyte, line 1047: This is kind of ugh to have this test code live here. Is there some way we can inject this into the Context struct when it's created?

src/context.rs
```diff
@@ -741,10 +1033,69 @@ impl DefaultSeafowlContext {
...
+
+        // We're not exposing `target_file_size` nor `write_batch_size` atm
+        let table = WriteBuilder::new()
+            .with_input_execution_plan(plan.clone())
+            .with_input_session_state(self.inner.state())
+            .with_object_store(table_object_store)
+            .await?;
+
+        // TODO: if `DeltaTable::get_version_timestamp` was globally public we could also pass the
+        // exact version timestamp, instead of creating onw automatically in our own catalog (which
```
mildbyte, line 1091:
```suggestion
        // exact version timestamp, instead of creating one automatically in our own catalog (which
```

src/context.rs
```diff
@@ -1391,8 +1675,15 @@ impl SeafowlContext for DefaultSeafowlContext {
                 // This is actually CREATE TABLE AS
                 let physical = self.create_physical_plan(input).await?;
 
-                self.execute_plan_to_table(&physical, None, Some(name.to_string()), None)
+                // First create the table and then insert the data from the subqeury
```
mildbyte, line 1678:
```suggestion
                // First create the table and then insert the data from the subquery
```

src/context.rs
```diff
@@ -1744,6 +2070,10 @@ impl SeafowlContext for DefaultSeafowlContext {
                                 .get_collection_id_by_name(&self.database, name)
                                 .await?
                             {
+                                // TODO: as for DROP TABLE, we should really only flag all tables
+                                // in the schema as DROPED, and then try to lazy delete them during
```
mildbyte, line 2074:
```suggestion
                                // in the schema as DROPPED, and then try to lazy delete them during
```

mildbyte (approved):
This is amazing and should let us close https://github.com/splitgraph/seafowl/issues/175 + delete a lot of our homegrown table version tracking code! Great work 👍

tests/data/seafowl-legacy-data/seafowl.sqlite-wal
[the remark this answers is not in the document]
gruuya: Oops, I just copied the folder over and proceeded to add the tests without really thinking about it.

I think I've seen this before; I'd prefer investigating after merging this PR if you agree.

src/system_tables.rs
gruuya: Oh this is just for our representation of the system table; the underlying catalog column defaults to `strftime('%s','now')` (for SQLite) and uses UTC.

I'll remove the comment as it's not really that important.

src/catalog.rs
gruuya: Yes there are—we use this UUID to scope our root object store for a particular table by adding it to the end of the path (or in other words, prefixing all table paths with it, relative to the root store uri). 

I'll think about a way to better handle no-column situation.

src/context.rs
gruuya: Good point, moving.

src/context.rs
gruuya: Good point, though I'm also inclined to cover this in a follow-up PR, since it's largely orthogonal to the delta-rs migration.

src/context.rs
gruuya: The alternatives that I see are:
1. Hide this behind a common indirection, e.g. just call `get_uuid()`, and then it would ultimately do the same thing as above.
2. Extend `DefaultSeafowlContext` with another field.

I feel that `DefaultSeafowlContext` struct is crammed enough already, so I dislike option #2 more than the current approach. Also, if we're bound to change the etag calculation logic this may not be necessary very soon, thus I'm inclined to keep it as is.

mildbyte (approved):

src/context.rs
mildbyte: Makes sense!

src/catalog.rs
gruuya: I realized that this comment is misleading—there's no way that the vec in `build_table` is empty since before we pass it down we group the original `collection_columns_vec` by table name (`.group_by(|col| &col.table_name)`), which means there was at least one element of the vector to begin with in order for grouping to work (likewise for `build_legacy_table`).

Moreover, in case of new tables, we only really need the uuid besides the name, which we can also fetch with the group by call as it's unique per table. In other words the rest of the `AllDatabaseColumnsResult` fields don't play a role in instantiating new tables, and I aim to simplify that loading logic after dropping the legacy tables completely.

On the other hand, this means that tables without columns wouldn't be loaded at all in the schema provider's map. However, this problem is averted since delta-rs doesn't allow column-less table creation (and we don't yet support `ALTER table DROP column`). Therefore, for now I'll just add a test to lock that in so it doesn't change underneath us.

Issue #323, opened 2023-03-22
Add ability to drop external tables


mildbyte:
Currently, it's impossible to drop staging external tables (they don't take any space but still hang out in the information_schema):

```bash
./examples/clients/node/seafowl-client.js "CREATE EXTERNAL TABLE tmp_pageeupqfplykfky7 STORED AS PARQUET LOCATION 'https://d37ci6vzurychx.cloudfront.net/trip-data/yellow_tripdata_2022-01.parquet';"
code: 200

./examples/clients/node/seafowl-client.js "DROP TABLE staging.tmp_pageeupqfplykfky7"
Error: 400: Error during planning: '"staging.tmp_pageeupqfplykfky7"' is a read-only table
    at IncomingMessage.<anonymous> (/Users/neumark/git/seafowl/examples/clients/node/seafowl-client.js:121:31)
    at IncomingMessage.emit (node:events:526:28)
    at emitCloseNT (node:internal/streams/destroy:138:10)
    at processTicksAndRejections (node:internal/process/task_queues:82:21)
```

gruuya:
Closed by #326

PR #326, opened 2023-03-23
Enable cached GET endpoint for authorized requests

### LINKED ISSUE #300: Allow authenticated cache-friendly query
[https://seafowl.io/docs/guides/querying-cache-cdn says](https://seafowl.io/docs/guides/querying-cache-cdn#note-on-authorization) says

> If you disable Seafowl reads or make them password-protected by setting read = "password"/"off", this will also disable the cached GET endpoint.

There should be no need for that. If you add `Authorization` to `Vary`, the cache should behave correctly with regard to different/missing authorization headers.

gruuya:
Consequently, also add the auth header to the VARY header. 

Also enable dropping of external tables.

Closes #300 and #323.


milesrichardson:
You might also consider adding the `Cache-Control: private` response header, which indicates that responses may be cached in local caches (e.g. the user's browser), but not in shared caches (e.g. Cloudflare). 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#response_directives

This would probably be more beneficial if Seafowl were using JWT for Authorization headers (since in that case, the `Vary` would Vary for the same user but different token). But it's also probably good practice in terms of security. 

I guess it's a design decision (maybe should be configurable?) whether you want intermediate caches like Cloudflare to cache private content.

src/frontend/http.rs
```diff
@@ -212,15 +212,24 @@ pub fn with_auth(
     )
 }
 
-// Disable the cached GET endpoint if we require auth for reads / they're disabled.
-// (since caching + auth is yet another can of worms)
+// Disable the cached GET endpoint if the reads are disabled. Otherwise extract the principle and
```
mildbyte, line 215:
```suggestion
// Disable the cached GET endpoint if the reads are disabled. Otherwise extract the principal and
```

milesrichardson:
Here's the [relevant Cloudflare documentation](https://developers.cloudflare.com/cache/about/cache-control/) for how they treat this feature, including some (kind of confusing) information on their defaults:

PR #341, opened 2023-04-04
Enable caching of S3 objects in the local FS

### LINKED ISSUE #41: Local object cache for S3

gruuya:
Also, enforce a TTL on the cache as a workaround for moka's eviction policy shortcomings for full caches:  https://github.com/moka-rs/moka/issues/147#issuecomment-1162899784

Closes #41.

### DIFF
--- a/src/config/schema.rs
+++ b/src/config/schema.rs
@@ -128,7 +128,7 @@ pub struct S3 {
 pub struct ObjectCacheProperties {
     pub capacity: u64,
     pub min_fetch_size: u64,
-    pub ttl: u64, // We could use humantime_serde crate to parse directly into `std::time::Duration`
+    pub ttl_s: u64, // We could use humantime_serde crate to parse directly into `std::time::Duration`
 }
 
 impl Default for ObjectCacheProperties {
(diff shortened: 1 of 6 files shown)


src/config/schema.rs
```diff
@@ -117,6 +120,25 @@ pub struct S3 {
...
     pub bucket: String,
+    pub cache_properties: Option<ObjectCacheProperties>,
+}
+
+#[derive(Deserialize, Debug, PartialEq, Eq, Clone)]
+#[serde(default)]
+pub struct ObjectCacheProperties {
+    pub capacity: u64,
+    pub min_fetch_size: u64,
+    pub ttl: u64, // We could use humantime_serde crate to parse directly into `std::time::Duration`
```
mildbyte, line 131: minor: could call it `ttl_s` to make it explicit that we expect a TTL in seconds.

mildbyte (approved):

PR #367, opened 2023-04-20
Return result type info on GET/POST endpoints

### LINKED ISSUE #356: Return result set type information in the HTTP response
We currently don't know the exact types of each column in the query result. We should encode the the information as JSON in a response header.

gruuya:
Provide this info via the content-type response header.

Closes #356.

### DIFF
--- a/src/frontend/http.rs
+++ b/src/frontend/http.rs
@@ -12,7 +12,7 @@ use arrow::json::LineDelimitedWriter;
 use arrow::record_batch::{RecordBatch, RecordBatchReader};
 use arrow_integration_test::{schema_from_json, schema_to_json};
 use arrow_schema::SchemaRef;
-use bytes::{BufMut, Bytes};
+use bytes::{Buf, Bytes};
 
 use datafusion::datasource::DefaultTableSource;
 use datafusion::parquet::arrow::arrow_reader::ParquetRecordBatchReader;
@@ -21,7 +21,7 @@ use datafusion_common::tree_node::{TreeNode, TreeNodeVisitor, VisitRecursion};
 use datafusion_expr::logical_plan::{LogicalPlan, TableScan};
 use deltalake::parquet::data_type::AsBytes;
 use deltalake::DeltaTable;
-use futures::{future, TryStreamExt};
+use futures::{future, StreamExt};
 use hex::encode;
 use log::{debug, info, warn};
 use percent_encoding::{percent_decode_str, utf8_percent_encode, NON_ALPHANUMERIC};
@@ -343,7 +343,7 @@ pub async fn upload(
     schema_name: String,
     table_name: String,
     user_context: UserContext,
-    form: FormData,
+    mut form: FormData,
     mut context: Arc<DefaultSeafowlContext>,
 ) -> Result<Response, ApiError> {
     if !user_context.can_perform_action(Action::Write) {
@@ -356,26 +356,22 @@ pub async fn upload(
         }
     }
 
-    let parts: Vec<Part> = form
-        .try_collect()
-        .await
-        .map_err(ApiError::UploadBodyLoadError)?;
-
     let mut has_header = true;
     let mut csv_schema: Option<Schema> = None;
     let mut filename = String::new();
-    for p in parts {
-        if p.name() == "has_header" {
-            let value_bytes =
-                load_part(p).await.map_err(ApiError::UploadBodyLoadError)?;
+    let mut part_data: Vec<u8> = vec![];
+    while let Some(maybe_part) = form.next().await {
+        let part = maybe_part.map_err(ApiError::UploadBodyLoadError)?;
+
+        if part.name() == "has_header" {
+            let value_bytes = load_part(part).await?;
 
             has_header = String::from_utf8(value_bytes)
                 .map_err(|_| ApiError::UploadHasHeaderParseError)?
                 .starts_with("true");
             debug!("Form part has_header is: {}", has_header);
-        } else if p.name() == "schema" && p.content_type() == Some("application/json") {
-            let value_bytes =
-                load_part(p).await.map_err(ApiError::UploadBodyLoadError)?;
+        } else if part.name() == "schema" {
+            let value_bytes = load_part(part).await?;
 
             csv_schema = Some(
                 schema_from_json(
@@ -384,39 +380,17 @@ pub async fn upload(
                 )
                 .map_err(ApiError::UploadSchemaParseError)?,
             );
-        } else if p.name() == "data" || p.name() == "file" {
-            filename = p.filename().ok_or(ApiError::UploadMissingFile)?.to_string();
+        } else if part.name() == "data" || part.name() == "file" {
+            filename = part
+                .filename()
+                .ok_or(ApiError::UploadMissingFile)?
+                .to_string();
 
             // Load the file content from the request
-            // TODO: we're actually buffering the entire file into memory here which is sub-optimal,
-            // since we could be writing the contents of the stream out into a temp file on the disk
-            // to have a smaller memory footprint. However, for this to be supported warp first
-            // needs to support streaming here: https://github.com/seanmonstar/warp/issues/323
-            let source = load_part(p).await.map_err(ApiError::UploadBodyLoadError)?;
-
-            // Parse the schema and load the file contents into a vector of record batches
-            let (schema, partition) =
-                match filename.split('.').last().ok_or_else(|| {
-                    ApiError::UploadMissingFilenameExtension(filename.clone())
-                })? {
-                    "csv" => load_csv_bytes(source, csv_schema.clone(), has_header)
-                        .map_err(|e| ApiError::UploadFileLoadError(e.into()))?,
-                    "parquet" => load_parquet_bytes(source)
-                        .map_err(ApiError::UploadFileLoadError)?,
-                    _ => return Err(ApiError::UploadUnsupportedFileFormat(filename)),
-                };
-
-            // Create an execution plan for yielding the record batches we just generated
-            let execution_plan = Arc::new(MemoryExec::try_new(
-                &[partition],
-                Arc::new(schema.clone()),
-                None,
-            )?);
-
-            // Execute the plan and persist objects as well as table/partition metadata
-            context
-                .plan_to_table(execution_plan, schema_name.clone(), table_name.clone())
-                .await?;
+            // TODO: we're actually buffering the entire file into memory here which is sub-optimal.
+            // Instead we should be writing the contents of the stream out into a temp file on the
+            // disk to have a smaller memory footprint.
+            part_data = load_part(part).await?;
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/frontend/http.rs
```diff
@@ -100,6 +102,25 @@ fn plan_to_etag(plan: &LogicalPlan) -> String {
...
 }
 
+// Construct a content-type header value that also includes schema information.
+fn content_type_with_schema(schema: SchemaRef) -> HeaderValue {
+    let output = schema
+        .fields
+        .iter()
+        .map(|f| format!("{}={}", f.name(), f.data_type()))
+        .collect::<Vec<String>>()
+        .join("; ");
```
gruuya, line 112: One alternative would be to use [serde serialization of the schema](https://github.com/apache/arrow-rs/blob/master/arrow-schema/src/schema.rs#L122) (so someone on the other side can get a proper arrow object), though I suspect that may end up being too convoluted.

src/frontend/http.rs
```diff
@@ -1548,12 +1599,24 @@ SELECT
...
+            resp.headers()
+                .get(header::CONTENT_TYPE)
+                .unwrap()
+                .to_str()
+                .unwrap(),
+            "application/json; smallint_val=Int16; integer_val=Int32; bigint_val=Int64; char_val=Utf8; \
+            varchar_val=Utf8; text_val=Utf8; float_val=Float32; real_val=Float32; double_val=Float64; \
+            bool_val=Boolean; date_val=Date32; timestamp_val=Timestamp(Nanosecond, None); \
+            int_array_val=List(Field { name: \"item\", data_type: Int64, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }); \
+            text_array_val=List(Field { name: \"item\", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} })",
```
milesrichardson, line 1618: Is there a name for this format of `Type(Params)`? Just wondering how a client might parse it. It looks almost like JSON but not quite.

gruuya: I don't think so, it's really only a serialization of a recursive type enum variant `DataType::List(Box<Field>)`. 

This however sort of breaks the abstraction, since we serialize only the name of the `Field` on the left, but on the right the entire (nested) `Field` struct gets unwrapped.

milesrichardson:
One thing I'd be particularly curious to test is whether Cloudflare caching behavior changes when including the `Content-Type` header, and if we can still use the trick of appending `.csv` to make sure it's cached.

src/frontend/http.rs
milesrichardson: Yeah, I guess the more human-readable verison would be `Int64[]`. Would you lose any information if you forced it to be like that? I guess `nullable` is important.

Related: What would be the type of a scalar column that is nullable? Is it even possible to produce such a query?

milesrichardson: I was able to construct such a query on my Seafowl test DB:

```
curl -i \
    -H "Content-Type: application/json" \
https://splitgraph-testing-nomad.seafowl.cloud/q -d@-<<EOF
{"query": "SELECT name, value, CASE when value > 2 THEN null WHEN value < 2 THEN 'ok' END islessthantwo FROM test_cache;"}
EOF
```

which returns the rows with nullable results as undefined (this is expected):

```
HTTP/2 200 
content-type: application/octet-stream
vary: Content-Type, Origin, X-Seafowl-Query
content-length: 386
date: Thu, 20 Apr 2023 14:54:11 GMT
server: Fly/ad79467d (2023-04-14)
via: 2 fly.io
fly-request-id: 01GYFJH6RQ0132ZX4000EK6RSY-lhr

{"name":"Alpha","value":2.3}
{"name":"Alpha","value":2.4}
{"name":"Beta","value":9.3}
{"name":"Beta","value":4.0}
{"name":"Gamma","value":6.8}
{"islessthantwo":"ok","name":"Gamma","value":1.2}
{"name":"Alpha","value":2.3}
{"name":"Alpha","value":2.4}
{"name":"Beta","value":9.3}
{"name":"Beta","value":4.0}
{"name":"Gamma","value":6.8}
{"islessthantwo":"ok","name":"Gamma","value":1.2}

```

So what would we expect the column type of `islessthantwo` to be in the header?

src/frontend/http.rs
```diff
@@ -1548,12 +1599,24 @@ SELECT
...
             .reply(&handler)
             .await;
         assert_eq!(resp.status(), StatusCode::OK);
+        assert_eq!(
+            resp.headers()
+                .get(header::CONTENT_TYPE)
+                .unwrap()
+                .to_str()
+                .unwrap(),
+            "application/json; smallint_val=Int16; integer_val=Int32; bigint_val=Int64; char_val=Utf8; \
```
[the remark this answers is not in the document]
milesrichardson: It's not a pain to parse on the client, because you can just `.split()` on `;` to get each parameter, and then for each parameter, split on `=` to get the `(key,value)`. The part that is a pain to parse is the types themselves, but that comes down to what we actually want to do with them.

However, I do agree that a JSON encoding of the schema would be easier to parse. It still doesn't solve the issue of what to call (how to interpret) the actual types like `List(...)` though.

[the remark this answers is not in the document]
mildbyte: >  you can just .split() on ; to get each parameter, and then for each parameter, split on = to get the (key,value)

what if my column name contains a semicolon? 🙃

src/frontend/http.rs
gruuya: > What would be the type of a scalar column that is nullable? Is it even possible to produce such a query?

Nullability is being [tracked at the `Field`](https://github.com/apache/arrow-rs/blob/master/arrow-schema/src/field.rs#L40) level, not the data type level, so in the case of a scalar column that is nullable there would be no info about it (as we'd record only the field name).

> There was some code that serializes column types into the database in the arrow-native JSON format if that didn't go away with the delta-rs migration.

That's still there, though it may be too [verbose](https://github.com/splitgraph/seafowl/blob/main/src/schema.rs#L72-L85).

> We could also serialize it into SQL datatypes (TEXT instead of Utf8, BIGINT[] instead of List(...Int64...))

I think this is the best approach overall, as it also keeps the output implementation agnostic, i.e. SQL in SQL out. I'll try to find a way to make this work.

milesrichardson:
I guess how we serialize it depends what we want to do with the data. It would be nice to use some standardized format, like a serialization of Arrow types, because then we could use (somewhat) standardized parsers if we want to coerce types (e.g. turning a DATETIME into a `Date()` object in JavaScript). However, if we're returning responses as JSON, and not a Parquet file, then what is the point of parsing it with Arrow?

I found arrow-js, which has code for coercing values to JavaScript types: https://github.com/apache/arrow/blob/main/js/src/type.ts

We could implement something like that in our client(s) (it's kind of up to the client what it does with it), or maybe somehow use arrow-js itself, but I think the important thing is that we would want the types to match the keys of this enum used by arrow:

https://github.com/apache/arrow/blob/main/js/src/enum.ts#L158-L203

Can we get those somehow? Is that what the serde serialization effectively does?

src/frontend/http.rs
mildbyte: We could also (which might be overkill) emit a JSONSchema instead of rolling our own custom format, which supports dates and floating point numbers. Could be useful for clients autogenerating parsers downstream.

milesrichardson:
It looks like arrow-js can maybe read a record batch from JSON? https://github.com/apache/arrow/blob/7ea2d9867cb90f9d0ca1db9b3bffa2fe8426e8ed/js/src/ipc/reader.ts#L153-L166

which calls

```
export const isArrowJSON = (x: any): x is ArrowJSONLike => {
    return isObject(x) && isObject(x['schema']);
};
```

but the `schema` type is just `any`, and I'm having trouble following how this works. Based on the arrow-python docs, it defaults to inferring types from values, but you can also pass it an explicit schema: https://arrow.apache.org/docs/python/json.html#automatic-type-inference

Interesting how it coerces type string by trying to convert it to a date first, and then a string.

----

This makes me think that it might be best to just use simple SQL types, since the motivating use case for this feature is displaying types to the user. And there is no avoiding that some coercion code needs to specify how to coerce values to special types (e.g. timestamp to `new Date()`), so it doesn't really matter what serialization format the header uses in that case, since the user needs to write some code to define which fields need to be coerced anyway (or a library can try to guess it based on the SQL type).

milesrichardson:
>  I wouldn't want us to pull in arrow-js (or copypaste code from it)

Yeah ofc. I was wondering if those types come from some arrow standard that we're already using. The arrow-js wouldn't be involved on the server side, but it would be nice if they were somehow compatible with some arrow-js code so that a client using arrow-json could load the HTTP jsonlines response into arrow from JSON and provide it with the schema from the header.

milesrichardson:
I think returning SQL-looking data types might be the simplest solution, but I worry it's non-standard (ofc SQL is standard, but it's the "-looking" part that concerns me).

When loading into arrow in JS, it seems unavoidable that the user needs to provide coercion functions: https://arrow.apache.org/docs/js/#create-a-table-from-javascript-arrays

The only exception would be if we were loading a `.arrow` file (Arrow IPC format) from a URL. So maybe in the future Seafowl could support returning data in Arrow IPC format. But as long as we're returning JSON, if we want to go back to arrow, it doesn't matter whether the schema is provided in "SQL-looking types" or arrow types, because we need to provide the coercion functions in either case. And it's easier to parse SQL-looking types, so I'd prefer those, I guess.

:shrug:

milesrichardson:
And we could always support returning both, defaulting to "SQL looking types," but the user could specify a preference in some request header (maybe `Content-Type`).

milesrichardson:
Ok, so I found the arrow-js code that can parse schema from JSON:

https://github.com/apache/arrow/blob/7ea2d9867cb90f9d0ca1db9b3bffa2fe8426e8ed/js/src/ipc/metadata/json.ts#L144-L206

Based on the fact it's accessing `["type"]["name"]`, I assume it expects the same serialization format that we see here in the nested types.

It would be cool to support this, because it means the JS client can use `arrow-js` and doesn't need to write its own parsers for the schema. But I don't intend to add that soon, it would just be a nice-to-have.

AFAICT, the "verbose serialization format" that @gruuya linked is what we would need for that: https://github.com/splitgraph/seafowl/blob/main/src/schema.rs#L72-L85 But it's of course quite verbose (are there length limits on this header?).

Meanwhile, for basic use cases like showing the user the type of the column in a shell, the SQL-looking types seem sufficient.

src/frontend/http.rs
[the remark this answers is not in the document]
milesrichardson: > what if my column name contains a semicolon? 🙃

I'm not sure this would be a valid HTTP parameter name, even if it was escaped. :/

src/frontend/http.rs
milesrichardson: It would definitely be nice to include nullability in this response header, even if it's technically part of `field.nullable` rather than `field.type`.

src/frontend/http.rs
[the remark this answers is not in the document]
gruuya: I didn't find any function/crate for converting Arrow type back to SQL-looking types, so I'd have to make a custom converter. In principle this would be fine for [simple types](https://github.com/apache/arrow-datafusion/blob/26bf31ebbc1274d0f2680d4b944145e794406188/datafusion/sql/src/planner.rs#L304), but for nested and recursive types it would get ugly. Plus, we'd still lose the nullability information regardless.

That's why I went ahead with using the [arrow schema JSON serialization format](https://arrow.apache.org/docs/format/Integration.html#json-test-data-format) now. Note that this is not an official format, it's just something that is used in integration testing, but is consistent. (Also note that I haven't employed url encoding for now.)

The returned schema is very verbose (check out the test assertions), though in theory analytical queries employ a smaller column cardinality (plus we're not "polluting" the actual result output as the schema info is hidden in the header). This could still pose a problem in terms of header size limits; as [per Cloudflare](https://developers.cloudflare.com/workers/platform/limits/#request-limits): `Request headers observe a total limit of 32 KB, but each header is limited to 16 KB`. Maybe we could make the schema type info in the headers optional, so that for some very large output column cardinality people can disable it.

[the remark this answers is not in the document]
gruuya: For the record, I've further modified this approach in order to url-encode the schema when sending via `content-type` header parameter, as discussed offline.

src/frontend/http.rs
```diff
@@ -100,6 +102,23 @@ fn plan_to_etag(plan: &LogicalPlan) -> String {
...
+    match HeaderValue::from_str(
+        format!("application/json; arrow-schema-escaped={output}").as_str(),
+    ) {
+        Ok(value) => value,
+        Err(err) => {
+            // Seems silly to error out here if the query itself succeeded.
+            warn!("Couldn't generate content type header for output schema {output}: {err:?}");
+            HeaderValue::from_static("application/json")
+        }
+    }
```
mildbyte, line 119: minor: [`unwrap_or_else`](https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap_or_else) could work here (no idea if you tried it; I haven't tested this code)

```suggestion
    HeaderValue::from_str(
        format!("application/json; arrow-schema-escaped={output}").as_str(),
    ).unwrap_or_else(|e| {
        // Seems silly to error out here if the query itself succeeded.
        warn!("Couldn't generate content type header for output schema {output}: {e:?}");
        HeaderValue::from_static("application/json")
    })
```

src/frontend/http.rs
```diff
@@ -100,6 +102,23 @@ fn plan_to_etag(plan: &LogicalPlan) -> String {
     encode(hasher.finalize())
 }
 
+// Construct a content-type header value that also includes schema information.
+fn content_type_with_schema(schema: SchemaRef) -> HeaderValue {
+    let schema_string = schema_to_json(schema.as_ref()).to_string();
+    let output = utf8_percent_encode(&schema_string, NON_ALPHANUMERIC);
+
+    match HeaderValue::from_str(
+        format!("application/json; arrow-schema-escaped={output}").as_str(),
```
mildbyte, line 111: I think we should still call it `arrow-schema`, even though it's percent encoded.

mildbyte (approved):

src/frontend/http.rs
gruuya: That's fine with me. @milesrichardson just an fyi that I'm revising the parameter name back to `arrow-schema`.

milesrichardson:
LGTM

PR #371, opened 2023-04-25
Bump object_store to 0.5.6

onpaws:
I originally expected we would have to fork `object_store` to benefit from `bcfe63ec5b0420b030e45091b8fc7c79e091646a` but upon digging deeper I learned it was already released, at least as of v0.5.6.

So this PR seeks to bring in that change and hopefully fix the GCS bug. Will post back findings 🤞

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1753,7 +1753,7 @@ dependencies = [
  "lazy_static",
  "log",
  "num_cpus",
- "object_store",
+ "object_store 0.5.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "parking_lot 0.12.1",
  "parquet",
  "percent-encoding",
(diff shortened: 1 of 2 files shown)


mildbyte:
Thanks! Mildly surprised that this is not reflected in Cargo.lock - did you forget to run `cargo update`?

image

> I originally expected we would have to fork object_store to benefit from bcfe63ec5b0420b030e45091b8fc7c79e091646a

You could also have pointed to a certain commit in the original arrow repo instead of forking it:

```toml
object_store = { git = "https://github.com/apache/arrow-rs", rev="bcfe63ec5b0420b030e45091b8fc7c79e091646a" }
```

or even (since the commit is on master)

```toml
object_store = { git = "https://github.com/apache/arrow-rs" }
```

But in any case, that probably wouldn't have patched transitive dependencies (e.g. the version of object_store that datafusion depends on), which is why we need to see what happens when the lock file is actually updated.

onpaws:
Closing in favor of 030d1086293becb4a4e6f54323e3231acd6d3af0
Thank you @gruuya

PR #379, opened 2023-04-28
GCS bucket support

onpaws:


src/config/schema.rs
```diff
@@ -54,6 +54,9 @@ pub enum ObjectStore {
     InMemory(InMemory),
     #[cfg(feature = "object-store-s3")]
     S3(S3),
+    #[cfg(feature = "object-store-gcs")]
+    #[serde(rename = "gcs")]
```
onpaws, line 58: I'm not sure why, but if I don't add this, we get this
image

neumark:
LGTM 👍

src/config/schema.rs
mildbyte: I think it's just how serde automatically names it and it looks more like an abbreviation to it, so it uses that format. What you did is fine.

src/object_store/wrapped.rs
```diff
@@ -42,6 +42,9 @@ impl InternalObjectStore {
             schema::ObjectStore::S3(S3 { bucket, .. }) => {
                 Url::from_str(&format!("s3://{bucket}")).unwrap()
             }
+            schema::ObjectStore::GCS(GCS { bucket, .. }) => {
+                Url::from_str(&format!("s3://{bucket}")).unwrap()
```
mildbyte, line 46: Sure this shouldn't be `gcs://`?

mildbyte (approved):
very nice, looking forward to the tests :)

src/config/schema.rs
onpaws: TIL serde has "abbreviation awareness." Cheers! 

Marking resolved.

src/config/context.rs
```diff
@@ -123,6 +124,26 @@ fn build_object_store(cfg: &schema::SeafowlConfig) -> Arc<dyn ObjectStore> {
...
+            bucket,
+            google_application_credentials,
+        }) => {
+            let gcs_builder: GoogleCloudStorageBuilder =
+                GoogleCloudStorageBuilder::new().with_bucket_name(bucket);
+
+            let gcs_builder = if let Some(path) = google_application_credentials {
+                gcs_builder.with_service_account_path(path)
+            } else {
+                gcs_builder
```
gruuya, line 140: It seems that this option [only works](https://github.com/apache/arrow-rs/blob/master/object_store/src/gcp/mod.rs#L1167-L1173) if Seafowl is running on GCP itself, such that the underlying client can [fetch the credentials](https://github.com/apache/arrow-rs/blob/master/object_store/src/gcp/credential.rs#L326-L333) from the metadata server.

This is fine in that case, though maybe we could just change the wording in the warning message then.

src/object_store/wrapped.rs
onpaws: It looks like according to [object_store](https://github.com/apache/arrow-rs/blob/eb5ac6953a9c24b8c5d03c95e4c49128b4e3b696/object_store/src/gcp/mod.rs#L942) the scheme should be `gs://`, so unless objections intending to go that route

fake-gcs-server.json
```diff
@@ -0,0 +1,6 @@
+{
+  "gcs_base_url": "http://localhost:4443",
+  "disable_oauth": true,
+  "client_email": "",
+  "private_key": ""
+}
```
onpaws, line 6: Minor:
It seems like better hygiene to move this little stub JSON cred into a "better" place. Any objections to e.g. `resources/`?

src/config/context.rs
onpaws: > we could just change the wording in the warning message then.
Thanks for the clarification! I [tweaked](https://github.com/splitgraph/seafowl/pull/379/files#diff-b1d0a32c21852f4c587c36f2e476a8966f37ce0973c58f04a28552f3645afbdaR355) the wording.

src/object_store/wrapped.rs
onpaws: Changed it to `gs://`. Marking resolved.

fake-gcs-server.json
gruuya: Yeah, moving that to `resources/` seems like a good idea. Even better if this credential can be passed via env vars/temp file like it's [done in delta-rs](https://github.com/delta-io/delta-rs/blob/main/rust/src/test_utils.rs#L40-L47).

Btw, given that XML multipart upload mock API seems to be [coming soon](https://github.com/fsouza/fake-gcs-server/issues/852#issuecomment-1493331946) to fake-gcs-server, I'd leave all the test code in this PR (but only remove the paramatrization from `test_insert_two_different_schemas` so that the tests pass for now).

gruuya (approved):

fake-gcs-server.json
onpaws: Thanks for the link to the `delta-rs` approach - I like it, arguably it's a bit nicer/cleaner to handle it that way vs leaving a single purpose stub file hanging around the repo.

RE the XML multipart upload, thanks for sharing that! 🤞 it lands soon, will be great to tie the knot on this

gruuya (approved):
Great, thanks!

PR #390, opened 2023-05-05
Remove legacy table reading logic

gruuya:
Also add migrations for dropping the obsolete metadata tables.

Note that the GC cleanup job now just removes the dropped table leftover files/folders, but it doesn't prune the table history.


migrations/postgres/20230505071708_remove_legacy_tables.up.sql
```diff
@@ -0,0 +1,3 @@
+DROP TABLE physical_partition_column;
```
mildbyte, line 1: Kinda scary to just delete these. How about, before we run the DB migrations, we check if the user still has some data in these tables and if so, output an error message ("make sure to use Seafowl 0.3 to run the Seafowl 0.3 -> 0.4 migration process") and exit (could even make a final final 0.3 release and do this migration process automatically)?

mildbyte (approved):
That's what they mean by no-code. LGTM!

migrations/postgres/20230505071708_remove_legacy_tables.up.sql
gruuya: Agreed, that makes sense. I've now added some checks to this effect (basically if there are any legacy tables or partitions we won't run 0.4 migrations), and a couple of tests as well.

PR #401, opened 2023-05-15
Add python Seafowl client library

neumark:


examples/clients/python/setup.py
```diff
@@ -0,0 +1,13 @@
+from distutils.core import setup
+
+setup(
+    name="python-seafowl-client",
+    version="0.0.1",
+    description="Python client for the Seafowl analytical SQL database.",
+    license="Apache 2.0",
+    author="Splitgraph, Inc",
+    author_email="hello@splitgraph.com",
+    url="https://github.com/matiascodesal/git-for-pip-example",
```
mildbyte, line 10: Wrong URL

examples/clients/python/seafowl/dataframe_to_sql.py
```diff
@@ -0,0 +1,83 @@
...
+    fields = ", ".join([f'"{sql_esc_ident(k)}"' for k in keys])
+    for row in data_iter:
+        sql = f'INSERT INTO "{sql_esc_ident(table.schema)}"."{sql_esc_ident(table.name)}" ({fields}) VALUES ({", ".join([emit_value(val) for val in row])});'
+        insert_row(sql)
+    return 0
+
+
+def execute_insert(
+    table: SQLTable, insert_row: RowInserter, chunksize: int | None = None
+) -> int | None:
```
mildbyte, line 57: Note this syntax is only supported in 3.10 and above. Given you still use `Optional` in other places, I propose using it here as well to keep compatibility with 3.9.

examples/clients/python/seafowl/dataframe_to_sql.py
```diff
@@ -0,0 +1,83 @@
...
+
+from .seafowl import DEFAULT_DB, DEFAULT_SCHEMA, emit_value, sql_esc_ident, query
+from .types import QualifiedTableName, RowInserter, SeafowlConnectionParams
+
+
+@contextmanager
+def with_pandas_sql() -> Generator[PandasSQL, None, None]:
+    engine = create_engine("sqlite://", echo=True)
+    with pandasSQL_builder(engine, need_transaction=False) as pandas_sql:
+        yield pandas_sql
```
mildbyte, line 19: What is this magic? 😅  Are we trying to get SQLAlchemy to emit Pandas -> SQL code in the SQLite dialect?

examples/clients/python/seafowl/dataframe_to_sql.py
```diff
@@ -0,0 +1,83 @@
...
+def get_table(
+    pandas_sql: PandasSQL, schema_name, table_name: str, data: pd.DataFrame
+) -> SQLTable:
+    return SQLTable(
+        table_name,
+        pandas_sql,
+        frame=data,
+        index=True,
+        # if_exists='replace',
+        # index_label=index_label,
```
mildbyte, line 31: Delete commented out code?

examples/clients/python/setup.py
```diff
@@ -0,0 +1,13 @@
...
+    packages=["seafowl"],
+    install_requires=["requests"],
```
mildbyte, line 12: Doesn't this need `sqlalchemy` (maybe we can avoid pulling it in) and `pandas`?

examples/clients/python/seafowl/dataframe_to_sql.py
neumark: Yes, exactly. SQLite and Seafowl have the same type names. The pandas `dtype` to SQL type conversion functions require a specific db adaptor, and we don't have one for Seafowl yet, but the SQLite one works like a charm -at least for generating a `CREATE TABLE ...` statement.

examples/clients/python/setup.py
neumark: 🤦 nice catch!

examples/clients/python/setup.py
```diff
@@ -0,0 +1,13 @@
+from distutils.core import setup
+
+setup(
+    name="python-seafowl-client",
```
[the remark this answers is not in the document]
neumark: Good point, I opted for `seafowl`

examples/clients/python/seafowl/seafowl.py
```diff
@@ -0,0 +1,83 @@
...
+    url = get_db_endpoint(conn.url, conn.database or DEFAULT_DB)
+    headers = {"Content-Type": "application/json"}
+    if conn.secret:
+        headers["Authorization"] = f"Bearer {conn.secret}"
+    response = requests.post(
+        f"{url}{URL_SUFFIX if append_url_suffix else ''}",
+        json={"query": sql},
+        headers=headers,
+    )
+    if raise_for_status and 400 <= response.status_code < 600:
```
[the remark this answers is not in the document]
neumark: Good point, I like delegating to `response.raise_for_status()` as you suggested!

examples/clients/python/setup.py
neumark: Not if you're just using `query()`.

mildbyte: Oh, I see, this won't crash if the user only imports `seafowl.seafowl`. Maybe add it as an extra?

mildbyte (approved):

PR #407, opened 2023-05-17
Make python CLI client print output similar to the node CLI client

neumark:


examples/clients/python/seafowl/__main__.py
```diff
@@ -40,10 +42,13 @@
...
+try:
+  status_code, data = query(
+      SeafowlConnectionParams(endpoint, password, None), sql, append_url_suffix=False
+  )
+  print(f"code: {status_code}")
+  if data:
+      print(json.dumps(data, indent=4))
+except HTTPError as error:
+    print(f"Error: {error.response.status_code}")
+    print(error.response.text)
```
mildbyte, line 54: I'd print this to stderr and exit with a non-zero code so that anything downstream knows this crashed.

mildbyte (approved):

examples/clients/python/seafowl/__main__.py
neumark: That's a good idea, updating the code accordingly!

PR #414, opened 2023-05-23
Support DataFrame uploads as Parquet in Python client

neumark:


examples/clients/python/seafowl/dataframe.py
```diff
@@ -63,7 +70,23 @@ def make_row_inserter(conn: SeafowlConnectionParams) -> RowInserter:
...
+def export_to_parquet_and_upload(
+    data: pd.DataFrame,
+    conn: SeafowlConnectionParams,
+    destination_table: QualifiedTableName,
+) -> None:
+    url = f"{get_db_endpoint(conn.url, conn.database)}/upload/{destination_table.schema or DEFAULT_SCHEMA}/{destination_table.table}"
+    with tempfile.NamedTemporaryFile(suffix=".parquet") as tmpfile:
+        data.to_parquet(tmpfile, compression="gzip")
+        tmpfile.flush()
+        tmpfile.seek(0)
```
gruuya, line 82: Just out of curiosity, is this `seek` needed? Will Seafowl re-use the same file handle?

gruuya (approved):

examples/clients/python/seafowl/dataframe.py
mildbyte: The `requests.post` uses this `tmpfile`, maybe it doesn't rewind automatically.

mildbyte (approved):

examples/clients/python/seafowl/dataframe.py
neumark: Right, I don't think it rewinds automatically.

gruuya: Understood, thanks!

PR #433, opened 2023-06-06
Enable creating external tables on cloud object stores

### LINKED ISSUE #256: feature: external table for private S3 bucket and ObjectStoreProvider
## Background

as of `0.2.10`, Seafowl is able to read external table with http(s) protocol, e.g.
- files in public S3 bucket
- pre-signed file in private S3 bucket

However, it fails to read private S3 files without pre-signed URL. E.g. register a S3 folder, which contains many files. File path could follow hive partitioned conventions.

Assume the seaflow config already use S3 object store in `bucketA` and the IAM user also have read access to another private bucket `bucketB`

```
[object_store]
type = "s3"
access_key_id = "******"
secret_access_key = "*"
region = "us-east-1"
bucket = "bucketA"
```

### Expect
```
curl -i -H "Content-Type: application/json" localhost:8080/q -H "Authorization: Bearer ******" -d@- <<EOF
{"query": "CREATE EXTERNAL TABLE tripdata \
STORED AS PARQUET \
LOCATION 's3://bucketB/yellow_tripdata_2022-10.parquet'
"}
EOF
```
HTTP/1.1 200

### Actual
HTTP/1.1 400 Bad Request

```
Internal error: No suitable object store found for s3://bucketB/yellow_tripdata_2022-10.parquet
```

gruuya:
This would allow one to query CSV/Parquet files stored in S3 and GCS. 

For example:
- query a different bucket with the same credentials as provided by in the config:
  ```bash
  $ curl -H "Content-Type: application/json" http://127.0.0.1:8080/q -d@- <<EOF
  {"query": "create external table test_aws stored as parquet location 's3://seafowl-public/tutorial/trase-supply-chains.parquet'; select * from staging.test_aws limit 1"}
  EOF
  {"commodity":"CORN","country_of_import":"CANADA","country_of_import_trase_id":"CA","country_of_production":"ARGENTINA","economic_bloc":"CANADA","exporter":"RONALB S R L","exporter_group":"RONALB S R L","exporter_group_id":63685,"exporter_id":39201,"exporter_trase_id":"AR-TRADER-3064104720","flow_id":401692173,"fob":24200.0,"importer":"RONALB S R L","is_domestic":"0","port":"ROSARIO","product_type":"CORN GRAINS","region_production_1_type":"COUNTRY","region_production_2":"SANTA FE","region_production_2_level":2,"region_production_2_trase_id":"AR-82","region_production_2_type":"PROVINCE","row_number":1,"scale":"SUBNATIONAL","version":"0.2.2","volume":50.0,"year":2018.0}
  ```
- query with new credential options (or just in another region):
  ```bash
  $ curl -H "Content-Type: application/json" http://127.0.0.1:8080/q -d@- <<EOF
  {"query": "create external table test_aws_options stored as parquet options ('access_key_id' '*******', 'secret_access_key' '*************', 'region' 'eu-west-3') location 's3://splitgraph-athena-test/supply-chains/supply-chains.parquet'; select * from staging.test_aws_options limit 1"}
  EOF
  {"commodity":"CORN","country_of_import":"CANADA","country_of_import_trase_id":"CA","country_of_production":"ARGENTINA","economic_bloc":"CANADA","exporter":"RONALB S R L","exporter_group":"RONALB S R L","exporter_group_id":63685,"exporter_id":39201,"exporter_trase_id":"AR-TRADER-3064104720","flow_id":401692173,"fob":24200.0,"importer":"RONALB S R L","is_domestic":"0","port":"ROSARIO","product_type":"CORN GRAINS","region_production_1_type":"COUNTRY","region_production_2":"SANTA FE","region_production_2_level":2,"region_production_2_trase_id":"AR-82","region_production_2_type":"PROVINCE","row_number":1,"scale":"SUBNATIONAL","version":"0.2.2","volume":50.0,"year":2018.0}
  ```

- GCS example (tested on a GCP VM, so the creds where automatically provided)
  ```bash
  $ curl -v -H "Content-Type: application/json" http://127.0.0.1:8080/q -d@- <<EOF
  {"query": "create external table test_gcs stored as parquet location 'gs://splitgraph-staging/tweets.parquet'; select * from staging.test_gcs limit 1"}
  EOF
  {"id":877940643162578944,"link":"https://www.twitter.com/AdamKinzinger/statuses/877940643162578944","screen_name":"AdamKinzinger","source":"Twitter for iPad","text":"Before we blame anyone else for the tone of today's politics, we must look to ourselves. I hope you'll join me.  https://t.co/O6WWDYloKl","time":"2017-06-22T13:25:35","user_id":18004222}
  ```

Closes #256

### DIFF
--- a/src/config/schema.rs
+++ b/src/config/schema.rs
@@ -1,18 +1,25 @@
+use std::collections::HashMap;
+use std::sync::Arc;
 use std::{
     fmt::{self, Display},
     path::{Path, PathBuf},
+    time::Duration,
 };
 
 use crate::object_store::cache::{
-    DEFAULT_CACHE_CAPACITY, DEFAULT_CACHE_ENTRY_TTL, DEFAULT_MIN_FETCH_SIZE,
+    CachingObjectStore, DEFAULT_CACHE_CAPACITY, DEFAULT_CACHE_ENTRY_TTL,
+    DEFAULT_MIN_FETCH_SIZE,
 };
 use config::{Config, ConfigError, Environment, File, FileFormat, Map};
+use datafusion_common::DataFusionError;
 use hex::encode;
 use log::{info, warn};
+use object_store::DynObjectStore;
 use rand::distributions::{Alphanumeric, DistString};
 use serde::Deserialize;
 use sha2::{Digest, Sha256};
 use sqlx::sqlite::SqliteJournalMode;
+use tempfile::TempDir;
 
 pub const DEFAULT_DATA_DIR: &str = "seafowl-data";
 pub const DEFAULT_SQLITE_DB: &str = "seafowl.sqlite";
@@ -126,10 +133,52 @@ pub struct S3 {
     pub cache_properties: Option<ObjectCacheProperties>,
 }
 
+impl S3 {
+    pub fn from_bucket_and_options(
+        bucket: String,
+        map: &HashMap<String, String>,
+    ) -> Result<Self, DataFusionError> {
+        Ok(S3 {
+            region: map.get("region").cloned(),
+            access_key_id: map
+                .get("access_key_id")
+                .ok_or(DataFusionError::Execution(
+                    "'access_key_id' not found in provided options".to_string(),
+                ))?
+                .clone(),
+            secret_access_key: map
+                .get("secret_access_key")
+                .ok_or(DataFusionError::Execution(
+                    "'secret_access_key' not found in provided options".to_string(),
+                ))?
+                .clone(),
+            endpoint: map.get("endpoint").cloned(),
+            bucket,
+            cache_properties: Some(ObjectCacheProperties::default()),
+        })
+    }
+}
+
 #[derive(Deserialize, Debug, PartialEq, Eq, Clone)]
 pub struct GCS {
     pub bucket: String,
     pub google_application_credentials: Option<String>,
+    pub cache_properties: Option<ObjectCacheProperties>,
+}
+
+impl GCS {
+    pub fn from_bucket_and_options(
+        bucket: String,
+        map: &HashMap<String, String>,
+    ) -> Self {
+        GCS {
+            bucket,
+            google_application_credentials: map
+                .get("google_application_credentials")
+                .cloned(),
+            cache_properties: Some(ObjectCacheProperties::default()),
+        }
+    }
 }
 
 #[derive(Deserialize, Debug, PartialEq, Eq, Clone)]
@@ -147,6 +196,21 @@ impl Default for ObjectCacheProperties {
             min_fetch_size: DEFAULT_MIN_FETCH_SIZE,
             ttl_s: DEFAULT_CACHE_ENTRY_TTL.as_secs(),
         }
+    }
+}
+
+impl ObjectCacheProperties {
+    pub fn wrap_store(&self, inner: Arc<DynObjectStore>) -> Arc<DynObjectStore> {
+        let tmp_dir = TempDir::new().unwrap();
+        let path = tmp_dir.into_path();
+
+        Arc::new(CachingObjectStore::new(
+            inner,
+            &path,
+            self.capacity,
+            self.min_fetch_size,
+            Duration::from_secs(self.ttl_s),
+        ))
     }
 }
 
--- a/src/context.rs
+++ b/src/context.rs
@@ -42,6 +42,7 @@ use std::sync::Arc;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use datafusion::common::DFSchema;
+use datafusion::datasource::listing::ListingTableUrl;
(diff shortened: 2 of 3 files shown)


src/config/schema.rs
```diff
@@ -126,10 +133,52 @@ pub struct S3 {
...
+                .get("access_key_id")
+                .ok_or(DataFusionError::Execution(
+                    "'access_key_id' not found in provided options".to_string(),
+                ))?
+                .clone(),
+            secret_access_key: map
+                .get("secret_access_key")
+                .ok_or(DataFusionError::Execution(
+                    "'secret_access_key' not found in provided options".to_string(),
+                ))?
```
gruuya, line 153: `DataFusionError` feels a bit foreign inside the config module.

src/context.rs
```diff
@@ -1170,6 +1174,47 @@ impl SeafowlContext for DefaultSeafowlContext {
...
+                                url.as_str()
+                            ))
+                        })?
+                        .to_string();
+
+                    let config = if scheme == "s3" {
+                        let s3 = if cmd.options.is_empty() && let schema::ObjectStore::S3(s3) = self.internal_object_store.config.clone() {
+                            S3{ bucket, ..s3 }
+                        } else {
+                            S3::from_bucket_and_options(bucket, &cmd.options)?
```
gruuya, line 1200: A potential alternative here is to use [get_s3_object_store_builder](https://docs.rs/datafusion-cli/latest/datafusion_cli/object_storage/fn.get_s3_object_store_builder.html) (and [get_s3_object_store_builder](https://docs.rs/datafusion-cli/latest/datafusion_cli/object_storage/fn.get_s3_object_store_builder.html) below, though that would mean adding a new dependency (datafusion-cli).

src/config/schema.rs
```diff
@@ -126,10 +133,52 @@ pub struct S3 {
...
+                .clone(),
+            endpoint: map.get("endpoint").cloned(),
+            bucket,
+            cache_properties: Some(ObjectCacheProperties::default()),
```
gruuya, line 157: Maybe we should be setting this to `None` in this case)

src/config/schema.rs
mildbyte: Yeah, we could define something ourselves or reuse a serde parsing error, given this is part of the config deserialization logic?

src/config/schema.rs
mildbyte: Is there a difference between `None` and `Some(ObjectCacheProperties::default())` (if someone doesn't want their object store contents to be cached, can they represent that in the config)? If not, we could replace the type with `ObjectCacheProperties` instead of the optional.

mildbyte (approved):

src/context.rs
gruuya: Sorry I meant for the other link to point to [get_gcs_object_store_builder](https://docs.rs/datafusion-cli/latest/datafusion_cli/object_storage/fn.get_gcs_object_store_builder.html).

In the end both end up doing effectively what we want—[create an object store dynamically and register it](https://docs.rs/datafusion-cli/latest/src/datafusion_cli/exec.rs.html#200-227) with the object store registry.

src/config/schema.rs
gruuya: Right; tbh I'm still torn torn between doing this here (or somewhere else in our code), or going with the datafusion-cli helper functions (which would make this redundant).

src/config/schema.rs
gruuya: > if someone doesn't want their object store contents to be cached, can they represent that in the config

this is currently represented by not specifying the `[object_store.cache_properties]` section in the config, and I think it's wise to leave it like that (i.e. keep it optional).

I'm just wondering whether it's worth it to setup the temp file caching machinery for a external table, and I'm leaning towards no for starters.

mildbyte: It could be useful. The reason is that we originally set up caching for the HTTP object store because DataFusion would call our object store requesting very small byte ranges very frequently, so we'd get slowed down by the latency (this could now have been fixed, but could still be a problem for S3/GCS).

src/config/schema.rs
gruuya: Looks like it's 61MB regardless we include `datafusion-cli` or not. I was also wary of using the df-cli functions as they initially source values from the env, and delta-rs used to dump options values to env vars when creating a new external table, but it seems this is no longer the case.

Given this, I eill try this approach next.

gruuya: Turns out delta-rs still dumps the env vars when creating an external delta table on S3, which would interfere with the creation of a new object store in [get_s3_object_store_builder](https://docs.rs/datafusion-cli/latest/datafusion_cli/object_storage/fn.get_s3_object_store_builder.html).

So I'll leave the implementation as is for now (only changing the error above and adding some tests).

Issue #357, opened 2023-04-12
Return timing data in response header(s)


milesrichardson:
Related: #356

**Feature Request**

It would be nice if the response from Seafowl included some information about execution time in the response headers. Currently, a client has no insight into execution metrics, other than measuring the total round trip time of the request and response.

**Motivation**

Having access to more timing data would be helpful for diagnostics, and for writing interactive query tools similar to the [Splitgraph Query Console](https://www.splitgraph.com/query). For example, the [Splitgraph DDN](https://www.splitgraph.com/connect/query) provides fields `executionTime` and `executionTimeHighRes`, which represent the time spent by the Splitgraph Engine "processing the query," which, in effect, is the interval defined by the boundaries where the "DDN web bridge" stops reading the HTTP request, and when it starts writing the HTTP response.

**Open questions:**

_**What should be measured?**_

* A minimal implementation should measure the total processing time, as delineated by the interval between when Seafowl finished reading the HTTP request body, and when it started writing the HTTP response body.
* Other more granular "internal" metrics might also be helpful, e.g. time spent resolving tables or fetching an external resource. But this generalizes to a telemetry framework, since it's possible to measure the time between any two events. So perhaps there are some important metrics worth including by default.

_**How should it be returned in the response headers?**_

All considerations from https://github.com/splitgraph/seafowl/issues/356#issuecomment-1505759519 apply:

* Each metric could be its own response header
* All metrics could be in one response header
* The most important metrics could be in the `parameters` of the `Content-Type` header, similarly to how we might return the field types. (The advantage of this option is that `Content-Type` is a default CORS-safelisted HTTP response header.)

For the `content-type` option, perhaps any non-field metadata could be in parameters conventionally prefixed with `__`, e.g.: 

```
content-type: application/octet-stream; __exec_ms=15; __resolving_ms=20; __total_ms=35; 
total_volume=FLOAT; year=INT;
...
{"total_volume":78.0,"year":2018}
```

onpaws:
OK, after flailing around in Warp but unf hitting my head straight into the (IMO confusing) world of `Filter`s, I ended up switching gears and now added an explicit timer inside `cached_read_query `/`uncached_read_write_query`. 
It's very simple and not yet DRY'd but wanted to push this up as a PoC of 'whole request' timing.

Probably whether Seafowl returns this header should either be 1) configurable or 2) perhaps it should follow in return only at INFO/DEBUG level. Or maybe both. (?)

Thoughts?

[...]

Issue #50, opened 2022-08-12
Expose some DataFusion runtime execution settings in the config


mildbyte:
- `MemoryManagerConfig`: max runtime memory usage for plan execution (rough, since it doesn't track basic process data structures): https://github.com/apache/arrow-datafusion/blob/master/datafusion/core/src/execution/memory_manager.rs#L35-L55 / https://github.com/apache/arrow-datafusion/blob/master/datafusion/core/src/execution/runtime_env.rs#L141-L145
- `DiskManagerConfig`: using the OS temp directory / other path to spill data to disk during execution: https://github.com/apache/arrow-datafusion/blob/master/datafusion/core/src/execution/disk_manager.rs#L50-L65 / https://github.com/apache/arrow-datafusion/blob/master/datafusion/core/src/execution/runtime_env.rs#L135-L139
- Planner/executor config options from https://github.com/apache/arrow-datafusion/blob/master/datafusion/core/src/config.rs and https://github.com/apache/arrow-datafusion/blob/master/datafusion/core/src/execution/context.rs#L1038-L1060

PR #438, opened 2023-06-15
Return timing data in response header(s)

onpaws:
Seeks to implement #357

- [x] Minimal implementation that seeks to measure "the interval between when Seafowl finished reading the HTTP request body, and when it started writing the HTTP response body".
- [x] Return that number inside `X-Seafowl-Runtime` (currently hardcoded)
- [x] Do we want the user to be able to customize (e.g. full control of header name and it's contents?) UPDATE 20230619: not at this time
- [x] Decide how that should be configurable e.g. via env var/`seafowl.toml`, or at the `INFO`/`DEBUG` level, or some other way
- [x] Add test for uncached flow
- [x] Add test for cached flow
- [x] Assuming we want to generalize this, consider which, if any existing standards [1][2] might make sense for Seafowl UPDATE 20230619: not intending to generalize in this PR, but in future maybe

[1] https://www.w3.org/TR/server-timing/
[2] https://arrow.apache.org/datafusion/user-guide/sql/explain.html


src/frontend/http.rs
```diff
@@ -53,6 +53,7 @@ const BEARER_PREFIX: &str = "Bearer ";
 // We have a very lax CORS on this, so we don't mind browsers
 // caching it for as long as possible.
 const CORS_MAXAGE: u32 = 86400;
+const RUNTIME_HEADER: &str = "X-Seafowl-Runtime";
```
gruuya, line 56: Runtime sounds weird to me in this context, maybe something like:

```suggestion
const QUERY_TIME_HEADER: &str = "X-Seafowl-Query-Time";
```

Also, there's the matter of CORS-safelisted headers that @milesrichardson raised in the original issue. I'm fine with having this info be in a separate header (i.e. not in `Content-Type`), but we'd probably need to specify [`Access-Control-Expose-Headers`](https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_response_header#extending_the_safelist) as well then.

gruuya:
Fwiw, I don't think this should be configurable (not for now at least); there are more pertinent knobs that still aren't customizable (e.g. see #50), and I'm wary of bloating up the config with less important items.

I think just extending the existing tests should be sufficient to merge this.

onpaws:
> Fwiw, I don't think this should be configurable (not for now at least); there are more pertinent knobs that still aren't customizable (e.g. see #50), and I'm wary of bloating up the config with less important items.
> 
> I think just extending the existing tests should be sufficient to merge this.

Thanks Marko, yes probably too soon to add new config options. In the future depending how we want to allow user customization e.g. once we return >1 timing measurement, it will presumably be time to consider.

I will switch to `X-Seafowl-Query-Time`, unless anyone objects, thanks!
In terms of answering "when should Seafowl return this header", it seems to me we have two options:

1. Always
2. Return only in certain conditions, e.g. if the user-agent passes e.g. `Seafowl-Trace: true` in the request

IIUC I think @mildbyte prefers option 1, and @milesrichardson prefers option 2. (at least, I had that impression based on MM chats.) 

What do you think @gruuya?

milesrichardson:
I would prefer option 1 (always) too. But just note that it won't always be available to client code in a cross-origin browsing context unless the header is included in the csv value of `Access-Control-Expose-Headers`.

But I'm hesitant to always return `Access-Control-Expose-Headers`. Perhaps *that* could be opt-in via a request flag. Or perhaps it's not a problem to always include it anyway, since any problem it would cause for someone running a proxy in front of seafowl, would only be relevant when they wanted the header, anyway - at which point they could add it to their own code where they specify allowed headers (that is, the relatively harmless failure case would be their own headers overriding a seafowl header that's meant to be opt-in anyway).

I had originally said that this shouldn't go into the `Content-Type` header alongside the schema. And it certainly feels like it would be overloading the semantic purpose of that header. But, after thinking more on it, I'm not convinced it's a problem to include timing data in addition to the schema. At first I thought it would be wrong because you don't want to cache it, but actually it makes sense to cache it, because it's timing how long it took to answer the original query, i.e. the one with the cached result. So it's totally sensible for the timing data to be included in the cached response that's served by Cloudflare, or even from your browser, since neither of those are re-executing the query, and the time it took when the query originally executed is still a relevant metric. 

So another question would be: what response headers does Cloudflare save in its cached objects? We know that it saves `Content-Type`, which is maybe a point in favor of including timing data there too. But I don't know what it does with other response headers (I assume it defaults to saving them?)

@gruuya @mildbyte What do you think about including the timing data in the `Content-Type` header along with the schema?

onpaws:
> I think just extending the existing tests should be sufficient to merge this.

Is [this](https://github.com/splitgraph/seafowl/pull/438/files#diff-51f622a727225cad9caed61cc103982c3a45ec6227624abe42a441d9c79d1ab3R1595) what you had in mind? Wasn't sure if you meant literally extending some already existing test, or adding to the existing test suite.

If there's some recommended way to hit both `cached_`.../`uncached_` paths from one single test, that seems nice to avoid what would likely otherwise be code duplication. 
I noticed certain macros e.g. [`#[case::uncached_post("POST", "/q")]`](https://github.com/splitgraph/seafowl/blob/c836955060655322decfe5b89cc279a2993d8302/src/frontend/http.rs#L1535) that suggest maybe we could use `uncached_get` and `cached_get` to test both paths from a single test? Not sure if I'm understanding that correctly though.
image

gruuya:
One thing I [just learned](https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header) is that apparently even default CORS safelisted headers have some prerequisites:
> For [Content-Type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type): needs to have a MIME type of its parsed value (ignoring parameters) of either application/x-www-form-urlencoded, multipart/form-data, or text/plain.

In our case we have [`application/json`](https://github.com/splitgraph/seafowl/blob/main/src/frontend/http.rs#L110-L117) so it seems we won't gain any benefit from shoehorning the timing data into `content-type` anyway.

gruuya:
@onpaws you could also simply extend the [last test](https://github.com/splitgraph/seafowl/blob/main/src/frontend/http.rs#L1521) in that module; it hits both GET and POSTS endpoints. In addition to presence, I'd test that the value of the header is as expected (i.e. something that can be parsed as a float, between 0 and 1).

milesrichardson:
@gruuya 

> One thing I [just learned](https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header) is that apparently even default CORS safelisted headers have some prerequisites:

Read that a bit more carefully: those restrictions are for the *request* `content-type` header and do not apply to the response. (I made the same mistake the first two times I read it in the original issue 😸)

gruuya:
Oh damn, good catch :D

onpaws:
> @onpaws you could also simply extend the [last test](https://github.com/splitgraph/seafowl/blob/main/src/frontend/http.rs#L1521) in that module; it hits both GET and POSTS endpoints. In addition to presence, I'd test that the value of the header is as expected (i.e. something that can be parsed as a float, between 0 and 1).

Ended up doing this approach, cheers.
Now we append it to the existing test hopefully in the way you desired. While I was at it I renamed it to `test_http_type_conversion_and_timing_header` which is perhaps a bit long but hopefully makes things clear for future us/contributors.

Agreed checking the header is "as expected" and added a second test for that. 
I wasn't initially sure the lines of responsibility/best ways to handle weird edge cases when parsing floats from strings. Along the way learned Rust core now uses [Eisel-Lemire](https://github.com/rust-lang/rust/pull/86761). Infinity and NaN scenarios were first a concern but IIUC it seems like [.is_finite()](https://github.com/splitgraph/seafowl/pull/438/commits/1bb617d860ee6a3f2f46b10d7152cdb0dc600844#diff-c351353ae7264ab39a8725d8657fa79580f379fa5aa81265cf152473e634de86R169) [handles both](https://doc.rust-lang.org/std/primitive.f64.html#method.is_finite), which I believe should cover us. If I missed any cases please lmk.

Verified the tests work by temporarily commenting out the header injection code, and re-running, and got the expected result.

<details>

<summary>before/after</summary>

the "before" test run, expected to fail b/c I commented out the header code
```
running 2 tests
test frontend::http::tests::test_http_type_conversion_and_timing_header::case_2_uncached_post ... FAILED
test frontend::http::tests::test_http_type_conversion_and_timing_header::case_1_cached_get ... FAILED

failures:

---- frontend::http::tests::test_http_type_conversion_and_timing_header::case_2_uncached_post stdout ----
thread 'frontend::http::tests::test_http_type_conversion_and_timing_header::case_2_uncached_post' panicked at 'assertion failed: resp.headers().contains_key(QUERY_TIME_HEADER)', src/frontend/http.rs:1596:9

---- frontend::http::tests::test_http_type_conversion_and_timing_header::case_1_cached_get stdout ----
thread 'frontend::http::tests::test_http_type_conversion_and_timing_header::case_1_cached_get' panicked at 'assertion failed: resp.headers().contains_key(QUERY_TIME_HEADER)', src/frontend/http.rs:1596:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    frontend::http::tests::test_http_type_conversion_and_timing_header::case_1_cached_get
    frontend::http::tests::test_http_type_conversion_and_timing_header::case_2_uncached_post

test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 188 filtered out; finished in 0.04s
```

the "after" run
```
running 2 tests
test frontend::http::tests::test_http_type_conversion_and_timing_header::case_2_uncached_post ... ok
test frontend::http::tests::test_http_type_conversion_and_timing_header::case_1_cached_get ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 188 filtered out; finished in 0.04s

     Running unittests src/main.rs (target/debug/deps/seafowl-49a42c42648f96f5)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/main.rs (target/debug/deps/main-084a7a60e84333ec)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 56 filtered out; finished in 0.00s

```

</details>

src/frontend/profile_utils.rs
```diff
@@ -0,0 +1,34 @@
...
+        Self { start_time: None }
+    }
+
+    pub fn start_timer(&mut self) {
+        self.start_time = Some(Instant::now())
+    }
+
+    pub fn elapsed(&self) -> Option<Duration> {
+        self.start_time.map(|start| start.elapsed())
+    }
```
Gelio, line 24: I'm not a Rust professional, so take this with a grain of salt.

It would be safer for the `start_time` field to be an `Instant`, without the `Option` wrapping. The timers added in this PR are started right after they are constructed. Thus, I don't see a point in splitting the `start_timer` method from the `new` constructor. We could avoid the unsafe `unwrap` at callsites of `elapsed` if we avoided `Option`s for `start_time`. If there is a timer that may not have started yet, we can use `Option<Timer>` rather than `Timer`+`start_time: Option<_>`

```suggestion
#[derive(Default)]
pub struct Timer {
    start_time: Instant,
}

impl Timer {
    pub fn new() -> Self {
        Self { start_time: Instant::now() }
    }

    pub fn elapsed(&self) -> Duration {
        self.start_time.elapsed()
    }
```

src/frontend/profile_utils.rs
```diff
@@ -0,0 +1,34 @@
+use std::time::{Duration, Instant};
+
+// Simple timer intended for profiling
+// Lazily inits. Returns `Duration`s or raw string (in ms)
+// Use Option as a safeguard because apparently previous Rust versions panicked
+// when current time was earlier than self.
```
Gelio, line 6: Minor:

[rustdoc comments should use triple-slashes](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html)

```suggestion
/// Simple timer intended for profiling
/// Lazily inits. Returns `Duration`s or raw string (in ms)
/// Use Option as a safeguard because apparently previous Rust versions panicked
/// when current time was earlier than self.
```

Gelio (commented):
Some minor comments from a fellow Rust enthusiast, feel free to ignore them, since they're mostly about the API of the newly added `Timer`

src/frontend/profile_utils.rs
gruuya: Agreed; also at this point it is questionable whether the `Timer` struct has any purpose, or should we skip the indirection and just use `Instant` directly in `uncached_read_write_query`/`cached_read_query`.

src/frontend/profile_utils.rs
onpaws: Done, thanks!

src/frontend/profile_utils.rs
gruuya: > Do you feel strongly I should drop Timer and use Instant directly?

If there's a need to accommodate some additional metrics (e.g. the ones from the plan) we can some fitting encapsulation later. For now it seems simpler to go with the bare `Instant`.

onpaws: Done, thanks!

onpaws:
CI passed and it seems we've probably reviewed this PR sufficiently. Any objections to merging?

onpaws:
> I haven't reviewed the code (and don't intend to), but once you merge it, can you update the docs in the marketing repo to reference the header?

Sure! How do you feel about e.g. this
image

with some basic `X-Seafowl-Query-Time` example similar to the other ones on this page?

gruuya (approved):

onpaws:
Thanks Marko! Merging now

PR #457, opened 2023-07-28
Support DROP FUNCTION

onpaws:
- [x] Add support `DROP FUNCTION [IF EXISTS] $myfunction`
- [x] Test it against 1 function
- [x] Test it against >1 function
- [x] Ensure error handling is where it needs to be
- [x] Add tests

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -111,7 +111,7 @@ use crate::{
     catalog::{FunctionCatalog, TableCatalog},
     data_types::DatabaseId,
     nodes::{
-        CreateFunction, CreateTable, DropSchema, RenameTable, SeafowlExtensionNode,
+        CreateFunction, CreateTable, DropFunction, DropSchema, RenameTable, SeafowlExtensionNode,
         Vacuum,
     },
     schema::Schema as SeafowlSchema,
@@ -1216,6 +1216,18 @@ impl SeafowlContext for DefaultSeafowlContext {
                         })),
                     }))
                 }
+                Statement::DropFunction{
+                    if_exists,
+                    func_desc,
+                    option
+                } => {
+                    Ok(LogicalPlan::Extension(Extension {
+                        node: Arc::new(SeafowlExtensionNode::DropFunction(DropFunction {
+                            func_desc,
+                            output_schema: Arc::new(DFSchema::empty()),
+                        }))
+                    }))
+                }
                 _ => Err(Error::NotImplemented(format!(
                     "Unsupported SQL statement: {s:?}"
                 ))),
@@ -1727,6 +1739,20 @@ impl SeafowlContext for DefaultSeafowlContext {
                                     name,
                                     *or_replace,
                                     details,
+                                )
+                                .await?;
+
+                            Ok(make_dummy_exec())
+                        }
+                        SeafowlExtensionNode::DropFunction(DropFunction {
+                            func_desc,
+                            output_schema: _,
+                         }) => {
+                            // Drop the function(s) from the metadata storage
+                            self.function_catalog
+                                .drop_function(
+                                    self.database_id,
+                                    func_desc,
                                 )
                                 .await?;
 
--- a/src/context.rs
+++ b/src/context.rs
@@ -2680,7 +2680,7 @@ mod tests {
 
         // Test dropping both functions in one pass
         let plan = sf_context
-            .plan_query(r#"DROP FUNCTION sintau, sintau"#)
+            .plan_query(r#"DROP FUNCTION sintau, sintau2"#)
             .await;
         assert!(plan.is_ok());
         Ok(())
--- a/src/context.rs
+++ b/src/context.rs
@@ -1752,7 +1752,6 @@ impl SeafowlContext for DefaultSeafowlContext {
                             self.function_catalog
                                 .drop_function(self.database_id, func_desc)
                                 .await?;
-
                             Ok(make_dummy_exec())
                         }
                         SeafowlExtensionNode::RenameTable(RenameTable {
@@ -2621,4 +2620,21 @@ mod tests {
             "Internal error: Error initializing WASM + MessagePack UDF \"invalidfn\": Internal(\"Error loading WASM module: failed to parse WebAssembly module"));
         Ok(())
     }
+
+    #[tokio::test]
+    #[ignore]
+    async fn test_drop_function() -> Result<()> {
+        let sf_context = in_memory_context().await;
+
+        // Source: https://gist.github.com/going-digital/02e46c44d89237c07bc99cd440ebfa43
+        let plan = sf_context
+            .plan_query(
+                r#"DROP FUNCTION nonexistentfunction"#,
+            )
+            .await;
+        assert!(plan.is_err());
+        Ok(())
+    }
+
+    // TODO we want to add a case for deleting an existing function
 }
--- a/src/context.rs
+++ b/src/context.rs
@@ -2642,12 +2642,12 @@ mod tests {
         let plan = sf_context
             .plan_query(r#"DROP FUNCTION IF EXISTS nonexistentfunction"#)
             .await;
-        assert!(!plan.is_err());
+        assert!(plan.is_ok());
         Ok(())
     }
 
     #[tokio::test]
-    async fn test_drop_two_functions() -> Result<()> {
+    async fn test_create_and_drop_two_functions() -> Result<()> {
         let sf_context = in_memory_context().await;
 
         let create_function_stmt = r#"CREATE FUNCTION sintau AS '
@@ -2676,11 +2676,7 @@ mod tests {
         let plan = sf_context
             .plan_query(r#"DROP FUNCTION sintau, sintau2"#)
             .await;
-        assert!(!plan.is_err());
+        assert!(plan.is_ok());
         Ok(())
     }
-
-
-
-    // TODO we want to add a case for deleting an existing function
 }
--- a/src/context.rs
+++ b/src/context.rs
@@ -111,8 +111,8 @@ use crate::{
     catalog::{FunctionCatalog, TableCatalog},
     data_types::DatabaseId,
     nodes::{
-        CreateFunction, CreateTable, DropFunction, DropSchema, RenameTable, SeafowlExtensionNode,
-        Vacuum,
+        CreateFunction, CreateTable, DropFunction, DropSchema, RenameTable,
+        SeafowlExtensionNode, Vacuum,
     },
     schema::Schema as SeafowlSchema,
     version::TableVersionProcessor,
@@ -1217,9 +1217,9 @@ impl SeafowlContext for DefaultSeafowlContext {
                     }))
                 }
                 Statement::DropFunction{
-                    if_exists,
+                    if_exists: _,
                     func_desc,
-                    option
+                    option: _
                 } => {
                     Ok(LogicalPlan::Extension(Extension {
                         node: Arc::new(SeafowlExtensionNode::DropFunction(DropFunction {
@@ -1747,13 +1747,10 @@ impl SeafowlContext for DefaultSeafowlContext {
                         SeafowlExtensionNode::DropFunction(DropFunction {
                             func_desc,
                             output_schema: _,
-                         }) => {
+                        }) => {
                             // Drop the function(s) from the metadata storage
                             self.function_catalog
-                                .drop_function(
-                                    self.database_id,
-                                    func_desc,
-                                )
+                                .drop_function(self.database_id, func_desc)
                                 .await?;
 
                             Ok(make_dummy_exec())
--- a/src/context.rs
+++ b/src/context.rs
@@ -1221,10 +1221,12 @@ impl SeafowlContext for DefaultSeafowlContext {
                     func_desc,
                     option: _
                 } => {
+                    let func_names: Vec<String> =
+                        func_desc.iter().map(|desc| desc.name.to_string()).collect();
                         Ok(LogicalPlan::Extension(Extension {
                             node: Arc::new(SeafowlExtensionNode::DropFunction(DropFunction {
                                 if_exists,
-                                func_desc,
+                                func_names,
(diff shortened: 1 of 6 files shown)


onpaws:
Tested: 
- [`CREATE FUNCTION`](https://github.com/splitgraph/seafowl-udf-go/blob/master/create_udf.sh)
```
./create_udf.sh
HTTP/1.1 200 OK
x-seafowl-query-time: 355
vary: Authorization, Content-Type, Origin, X-Seafowl-Query
transfer-encoding: chunked
date: Fri, 28 Jul 2023 21:39:26 GMT
```
image

- [`DROP FUNCTION`](https://github.com/splitgraph/seafowl-udf-go/blob/master/delete_udf.sh)
```
./delete_udf.sh
HTTP/1.1 200 OK
x-seafowl-query-time: 388
vary: Authorization, Content-Type, Origin, X-Seafowl-Query
transfer-encoding: chunked
date: Fri, 28 Jul 2023 21:38:48 GMT
```
image

src/repository/default.rs
```diff
@@ -439,6 +439,54 @@ impl Repository for $repo {
...
+            // Execute the SQL DELETE query 1x per function
+            let mut query = sqlx::query(&query)
+                .bind(database_id)
+                .bind(function_name);
+
+            if let Some(_args) = args.clone() {
+                query = query.bind(args);
+            }
+
+            query
```
onpaws, line 481: In SQL, running `DROP FUNCTION` for a non-existent function, e.g. running it 2x in a row, will error:

https://github.com/splitgraph/seafowl/assets/182515/041897cd-8933-4843-97a5-73b96b5412a5

Conversely, `DELETE FROM`-ing a table twice in a row does not:

https://github.com/splitgraph/seafowl/assets/182515/ee20ec8b-5fa6-438b-995b-e4ea6fac9a40


This is normally fine, but because I implemented `DROP FUNCTION` using `DELETE FROM`, running `DROP FUNCTION` 2x in a row does _not_ currently yield some kind of error, e.g. `Error::Plan`, though it seems like it probably should.

One idea I had was when we execute DELETE FROM, tally up `rows_affected` and if > 0, return Err(). 
```rust
  let rows_affected = query
                .execute(&self.executor)
                .await
                .map_err($repo::interpret_error)?
                .rows_affected();

            if rows_affected == 0 {
                return Err($SOME_ERROR);
            }
```

However, I ran into trouble importing [FunctionDoesNotExist](https://github.com/splitgraph/seafowl/pull/457/files#diff-99aeb2649cc42795a63369ffb1ba0f206d9f24bb0779b095f01e9b687481064eR48)

```rust
if rows_affected == 0 {
    return Err(crate::catalog::Error::FunctionDoesNotExist { name: function_name.to_string() });
}
```

```
error[E0308]: mismatched types
   --> src/repository/default.rs:488:28
    |
488 |                 return Err(crate::catalog::Error::FunctionDoesNotExist { name: function_name.to_string() });
    |                        --- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `interface::Error`, found `catalog::Error`
    |                        |
    |                        arguments to this enum variant are incorrect

```

Trying to figure out an appropriate way to handle. Thoughts?

gruuya: > running DROP FUNCTION 2x in a row does not currently yield some kind of error, e.g. Error::Plan, though it seems like it probably should.

It should, conditional on the user omitting [`IF EXISTS` clause](https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/ast/mod.rs#L1407) in the `DropFunction` statement. If that clause is used then the statement shouldn't error out in either case. Here's an analogous example: https://github.com/splitgraph/seafowl/blob/main/src/context.rs#L1353-L1359

> One idea I had was when we execute DELETE FROM, tally up rows_affected and if > 0, return Err().

You're on the right track; what we usually do is add a `RETURNING id` to the `DELETE` statement, so that the DB nicely handles that check for us: https://github.com/splitgraph/seafowl/blob/main/src/repository/default.rs#L444-L454

The reason for the error above is that you're doing that check manually inside the `Repository` trait implementation which expects an [interface error](https://github.com/splitgraph/seafowl/blob/main/src/repository/interface.rs#L80-L86), while you specify the [catalog error](https://github.com/splitgraph/seafowl/blob/main/src/catalog.rs#L38-L51). By following the above prescription you can circumvent this issue.

src/context.rs
```diff
@@ -2598,4 +2620,19 @@ mod tests {
...
+    #[tokio::test]
+    #[ignore]
+    async fn test_drop_function() -> Result<()> {
+        let sf_context = in_memory_context().await;
+
+        // Source: https://gist.github.com/going-digital/02e46c44d89237c07bc99cd440ebfa43
+        let plan = sf_context
+            .plan_query(r#"DROP FUNCTION nonexistentfunction"#)
+            .await;
+        assert!(plan.is_err());
```
gruuya, line 2633: You could also match on the error message.

(also the comment above is obsolete.)

src/repository/default.rs
```diff
@@ -439,6 +439,54 @@ impl Repository for $repo {
         Ok(functions)
     }
 
+
+    async fn drop_function(
+        &self, database_id:
+        DatabaseId,
+        func_desc: &[sqlparser::ast::DropFunctionDesc],
```
gruuya, line 446: A `sqlparser` struct all the way down in our repo implementation feels out of place. We should probably parse the required info during logical planning and populate adequate fields in our extension node at that point.

On that note, given that the function name is a unique key in our case (in [combination with the DB id](https://github.com/splitgraph/seafowl/blob/main/migrations/postgres/20220705072248_create_tables.up.sql#L68)), there's no need to require/keep track of the args, only the names supplied in the statement.

src/repository/default.rs
onpaws: Thank you Marko! If I understand you correctly `if_exists` will be true, IFF the user provided `IF EXISTS` in their query. I'm adapting the rest of the logic accordingly now, thanks.

src/repository/default.rs
onpaws: Thanks for the pointer to the DB id definition. 
Given the function names provide sufficient uniqueness it seems like `func_names: Vec<String>` should be safe. I went ahead and did this in the most recent commit. 
Going to mark resolved, but if I've made an error please feel free to un-resolve/comment as needed.

src/context.rs
onpaws: Are you saying to match on the string contents?

```
$ ./delete_udf.sh
HTTP/1.1 400 Bad Request
vary: Authorization, Content-Type, Origin, X-Seafowl-Query
content-length: 230
date: Sat, 05 Aug 2023 07:11:59 GMT

Internal error: Internal SQL error: "no rows returned by a query that expected to return at least one row". This was likely caused by a bug in DataFusion's code and we would welcome that you file an bug report in our issue tracker%                    
```

e.g. 'no rows returned'? 

I had the impression `plan.is_err()` might be sufficient to match on, but not 100% sure the pros/cons - would be curious your thoughts.

src/repository/default.rs
onpaws: OK, I believe these changes are implemented now. Going to mark resolved, but if any errors/changes needed, please feel free to re-open.

src/context.rs
```diff
@@ -1216,6 +1216,19 @@ impl SeafowlContext for DefaultSeafowlContext {
...
                 }
+                Statement::DropFunction{
+                    if_exists,
+                    func_desc,
+                    option: _
+                } => {
+                        Ok(LogicalPlan::Extension(Extension {
+                            node: Arc::new(SeafowlExtensionNode::DropFunction(DropFunction {
+                                if_exists,
+                                func_desc,
```
gruuya, line 1227: You could also extract those function names all the way up here (and change the corresponding field in `SeafowlExtensionNode::DropFunction` to be `func_names: Vec<String>`).

src/repository/default.rs
```diff
@@ -439,6 +439,51 @@ impl Repository for $repo {
...
+                    RETURNING id;
+                    "#,
+                );
+
+            // Execute the SQL DELETE query 1x per function
+            let query = sqlx::query(&query)
+                .bind(database_id)
+                .bind(function_name);
+
+            println!("database_id: {}, function_name: {}", database_id, function_name);
```
gruuya, line 466: `debug!` if you want to keep the log?

src/repository/default.rs
```diff
@@ -439,6 +439,51 @@ impl Repository for $repo {
...
+    ) -> Result<(), Error> {
+
+        for function_name in func_names.iter() {
+            // Construct the SQL DELETE statement
+            let query =
+                format!(
+                    r#"
+                    DELETE FROM "function"
+                    WHERE database_id = $1 AND name = $2
+                    RETURNING id;
```
gruuya, line 457: You can delete all functions with one SQL statement: `DELETE FROM ... AND name IN (...). To construct that query you can use the sqlx query builder (see for example `get_all_table_versions`).

src/context.rs
gruuya: The idea is that the error we hit is also the one we expect. The one above is directly from sqlite/postgres so we'd need to translate it to match the specific case being handled here i.e. function didn't exist in the first place.

I think you'd need to match for `Err(RepositoryError::SqlxError(sqlx::error::Error::RowNotFound))` in the catalog `drop_function` call, and then if `if_exists` is true just swallow that error, otherwise surface it with a custom message.

src/repository/default.rs
```diff
@@ -439,6 +439,51 @@ impl Repository for $repo {
...
+                    .map_err($repo::interpret_error)?;
+            } else {
+                // Otherwise if no function found, we return error
+                query
+                    .fetch_one(&self.executor)
+                    .await
+                    .map_err($repo::interpret_error)?;
+            }
+
+        }
```
gruuya, line 482: If you go for handling/converting `Err(RepositoryError::SqlxError(sqlx::error::Error::RowNotFound))` in the catalog, then you can simplify this to one expression (using `fetch_one`), and there's no need to pass `if_exists` down here.

src/repository/default.rs
onpaws: Good idea, did it that way!

src/repository/default.rs
onpaws: Dropped it, cheers.

src/context.rs
onpaws: Done 👍

src/repository/default.rs
onpaws: Done 👍

src/context.rs
onpaws: Now that we restored `FunctionNotFound` inside catalog.rs, I updated the tests to specifically check for the string literal. Marking resolved, please re-open as needed

gruuya (approved):

gruuya:
Very nice!

onpaws:
Unless objections, intending to merge this later today, and also the marketing/docs PR

@mildbyte OK to push the button? Anything need attention?

src/context.rs
```diff
@@ -2598,4 +2623,67 @@ mod tests {
...
+        );
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_drop_function_if_exists() -> Result<()> {
+        let sf_context = in_memory_context().await;
+
+        // Source: https://gist.github.com/going-digital/02e46c44d89237c07bc99cd440ebfa43
```
mildbyte, line 2648: redundant? (some wasm stuff) 

```suggestion
```

src/context.rs
```diff
@@ -2598,4 +2623,67 @@ mod tests {
...
+            "data": "AGFzbQEAAAABDQJgAX0BfWADfX9/AX0DBQQAAAABBQQBAUREBxgDBnNpbnRhdQAABGV4cDIAAQRsb2cyAAIKjgEEKQECfUMAAAA/IgIgACAAjpMiACACk4siAZMgAZZBAEEYEAMgAiAAk5gLGQAgACAAjiIAk0EYQSwQA7wgAKhBF3RqvgslAQF/IAC8IgFBF3ZB/wBrsiABQQl0s0MAAIBPlUEsQcQAEAOSCyIBAX0DQCADIACUIAEqAgCSIQMgAUEEaiIBIAJrDQALIAMLC0oBAEEAC0Q/x2FC2eATQUuqKsJzsqY9QAHJQH6V0DZv+V88kPJTPSJndz6sZjE/HQCAP/clMD0D/T++F6bRPkzcNL/Tgrg//IiKNwBqBG5hbWUBHwQABnNpbnRhdQEEZXhwMgIEbG9nMgMIZXZhbHBvbHkCNwQAAwABeAECeDECBGhhbGYBAQABeAICAAF4AQJ4aQMEAAF4AQVzdGFydAIDZW5kAwZyZXN1bHQDCQEDAQAEbG9vcA=="
+        }';"#;
+
+        // Create two functions in two separate passes
+        sf_context.plan_query(create_function_stmt).await?;
+        sf_context.plan_query(create_function_stmt2).await?;
+
+        // Test dropping both functions in one pass
+        let plan = sf_context
+            .plan_query(r#"DROP FUNCTION sintau, sintau2"#)
```
mildbyte, line 2684: What happens if you do `DROP FUNCTION sintau, sintau`? 🙃 (I think it just runs a SQL `DELETE ... WHERE function_name IN ('sintau', 'sintau')` and it'll be fine)

mildbyte (approved):

mildbyte:
@onpaws Just a small cleanup / nitpick. LGTM, thanks!

src/context.rs
onpaws: Thanks! Indeed, just dropped it

src/context.rs
onpaws: Good shout. Just applied that edit
image
and invoked the test
image
and it still passes. 

So yeah, I think you're right it's fine.

PR #460, opened 2023-08-14
Enable time travel for write statements with source queries

### LINKED ISSUE #170: Extend support for table time travel syntax to write statements
The [current table time travel](#154) syntax doesn't support versioning in write subqueries (e.g. `CREATE TABLE old_table AS SELECT * FROM new_table('2022-01-01 10:10:10 +00:00')`), and instead the table ref gets resolved to the latest table version. The reason is that we only do AST rewrites for the Query (aka SELECT) root node. For this to work we'd need to extend the visitor for more AST root nodes (CREATE, INSERT at least).

gruuya:
Also moved the existing and the new time travel test into a separate file.

Closes #170.


src/context.rs
```diff
@@ -933,57 +986,24 @@ impl SeafowlContext for DefaultSeafowlContext {
...
+        // Create a mutable clone of the statement so that we can rewrite table names if we encounter
+        // time travel syntax.
+        // Alternatively, this could be done without the `mut`, except then we'd need to construct
+        // and return a new `DFStatement` after the rewrite. In case of `Statement::Query` this is
+        // straight-forward as that enum variant contains a struct, however `Statement::Insert` and
+        // `Statement::CreateTable` wrap dozens of fields in curly braces, and we'd need to capture
+        // and re-populate each of those, so it would be very verbose.
+        // TODO: If `sqlparser-rs` encapsulates these fields inside a struct at some point remove
+        // the mut and go with @ capture and de-structuring to pass along other fields unchanged.
+        let mut stmt = statement.clone();
```
gruuya, line 999: We already `clone`d the statement so that wont incur a time penalty. In theory the `mut` might, but I think it's negligible if any, given that it's local to a single function.

The alternative I mention is to have things as before and then capture and re-create a enum value with all the fields explicitly (since @ capture doesn't work in this case), i.e. https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/ast/mod.rs#L1178-L1198 for insert (even more verbose for create table).

mildbyte (approved):

PR #472, opened 2023-11-24
Support for sub-directories in object stores

### LINKED ISSUE #470: Add support for object stores with specific paths
Currently we only support storing data rooted at the object store's bucket level.

In principle users could supply a custom path on the object store where they'd like Seafowl to store data.

This would also allow #469 to for instance have a staging area for parquet tables in one path/bucket and the actual Seafowl data dir in another and then convert the former into proper Delta tables.

gruuya:
This will allow configuring a specific path alongside a bucket where to store the data in.

Closes #470.

### DIFF
--- a/src/config/context.rs
+++ b/src/config/context.rs
@@ -83,9 +83,8 @@ pub fn build_object_store(
             endpoint,
             bucket,
             cache_properties,
+            ..
         }) => {
-            // Bucket from the config may contain some path so extract the proper bucket
-            let bucket = bucket.split('/').next().unwrap_or(bucket);
             let mut builder = AmazonS3Builder::new()
                 .with_access_key_id(access_key_id)
                 .with_secret_access_key(secret_access_key)
@@ -110,9 +109,8 @@ pub fn build_object_store(
             bucket,
             google_application_credentials,
             cache_properties,
+            ..
         }) => {
-            // Bucket from the config may contain some path so extract the proper bucket
-            let bucket = bucket.split('/').next().unwrap_or(bucket);
             let gcs_builder: GoogleCloudStorageBuilder =
                 GoogleCloudStorageBuilder::new().with_bucket_name(bucket);
 
--- a/src/config/context.rs
+++ b/src/config/context.rs
@@ -9,7 +9,7 @@ use crate::{
 };
 use datafusion::execution::context::SessionState;
 use datafusion::{
-    error::DataFusionError,
+    common::Result,
     execution::runtime_env::{RuntimeConfig, RuntimeEnv},
     prelude::{SessionConfig, SessionContext},
 };
@@ -69,12 +69,11 @@ async fn build_catalog(
 
 pub fn build_object_store(
     object_store_cfg: &schema::ObjectStore,
-) -> Arc<dyn ObjectStore> {
-    match &object_store_cfg {
-        schema::ObjectStore::Local(schema::Local { data_dir }) => Arc::new(
-            LocalFileSystem::new_with_prefix(data_dir)
-                .expect("Error creating object store"),
-        ),
+) -> Result<Arc<dyn ObjectStore>> {
+    Ok(match &object_store_cfg {
+        schema::ObjectStore::Local(schema::Local { data_dir }) => {
+            Arc::new(LocalFileSystem::new_with_prefix(data_dir)?)
+        }
         schema::ObjectStore::InMemory(_) => Arc::new(InMemory::new()),
         #[cfg(feature = "object-store-s3")]
         schema::ObjectStore::S3(S3 {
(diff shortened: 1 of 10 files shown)


src/config/context.rs
```diff
@@ -85,6 +84,8 @@ pub fn build_object_store(
             bucket,
             cache_properties,
         }) => {
+            // Bucket from the config may contain some path so extract the proper bucket
+            let bucket = bucket.split('/').next().unwrap_or(bucket);
```
gruuya, line 88: I went with overloading the bucket config param so that it can also hold a path to stay backward-compatible.

But I think the preferable way to do this is to ditch the `type` and `bucket` fields and replace them with a `location` field which can store all the relevant info (i.e. type, bucket and optionally path) at the expense of breaking backward-compatibility.

mildbyte: Could keep the `bucket` and have a separate `prefix` field that is optional and defaults to no prefix where it doesn't exist?

src/object_store/wrapped.rs
```diff
@@ -252,3 +287,63 @@ impl ObjectStore for InternalObjectStore {
...
+        let config = ObjectStore::S3(S3 {
+            region: None,
+            access_key_id: "access_key_id".to_string(),
+            secret_access_key: "secret_access_key".to_string(),
+            bucket: bucket.to_string(),
+            endpoint: endpoint.clone(),
+            cache_properties: None,
+        });
+        // In principle for this test we could use any object store since we only exercise the
+        // prefix/log store uri logic
```
mildbyte, line 328: If you're worried about this, you could break out `table_prefix` into a separate function that operates on `(schema::ObjectStore, Url) -> Url` and test it without having to create an object store (don't know how much it's worth it).

src/object_store/wrapped.rs
```diff
@@ -33,51 +33,86 @@ pub struct InternalObjectStore {
...
+                let mut segments =
+                    self.root_uri.path_segments().unwrap().collect::<Vec<_>>();
+
+                // Add the last path segment which is the table UUID
+                let uuid_str = table_uuid.to_string();
+                segments.push(&uuid_str);
+
+                // Remove the first path segment which is actually the bucket
+                segments.remove(0);
+                Path::from(segments.join("/"))
```
mildbyte, line 87: Maybe having a `prefix` as a separate config field could simplify this logic or avoid having to do this on every query.

tests/statements/ddl.rs
```diff
@@ -275,7 +284,12 @@ async fn test_create_table_move_and_drop(
 #[rstest]
 #[tokio::test]
 async fn test_create_table_drop_schema(
-    #[values(ObjectStoreType::InMemory, ObjectStoreType::Local, ObjectStoreType::S3)]
+    #[values(
+        ObjectStoreType::InMemory,
+        ObjectStoreType::Local,
+        ObjectStoreType::S3(None),
+        ObjectStoreType::S3(Some("/path/to/folder"))
```
mildbyte, line 291: Does this work with or without the leading slash?

mildbyte (approved):

tests/statements/ddl.rs
gruuya: Yup, works in both cases, the `test_table_location_s3` actually covers that.

src/config/context.rs
gruuya: I thought that a separate `prefix` field wasn't worth the increased config complexity, whereas the `location` approach felt more consolidated than either alternative.

To me it seems like it's natural to think about the full url `location` as one thing from the user perspective, and leave the parsing of the individual details (type, bucket, path) to Seafowl. This would also neatly get rid of the `endpoint` field on S3, as this is covered by `location` too. Moreover, currently we don't support passing the bucket/path in the `endpoint`, though in principle there's no reason not to. [GlareDB uses](https://github.com/GlareDB/glaredb/blob/b9cea63d1326e4528be73bd154cd13f16637302d/.github/workflows/ci.yaml#L315-L319) the `location` + storage options map approach and does away with config file completely due to the conciseness of this approach, though this wouldn't work for Seafowl as we have a bunch of other configurations as well.

To clarify, I definitely do prefer a separate `path` (or `prefix`) field over overloading the `bucket` field, I just think `location` would be even better from a usability standpoint. Let me know what are your preferences.

src/object_store/wrapped.rs
gruuya: Not really worried, intended more like a clarification. The test also exercises the log store uri, which does need some object store.

src/object_store/wrapped.rs
gruuya: Yes, in this case it would.

src/config/context.rs
gruuya: Alright, I made the change to use the `prefix` field for now as that's good enough for a merge I believe, but we can revisit whether it makes sense to go with the `location` approach later on.

PR #473, opened 2023-11-27
Convert to Delta table support

### LINKED ISSUE #469: Implement conversion of parquet to Delta tables
Introduce an analogue to https://docs.databricks.com/en/sql/language-manual/delta-convert-to-delta.html

This should make use of https://github.com/delta-io/delta-rs/pull/1686 behind the scenes. It should also work in-place (i.e. for parquet tables stored inside the Seafowl data dir) and optionally for parquet tables stored in other folders/buckets or even object stores.

gruuya:
Convert a parquet table, as specified by a particular path, into a Delta table. Syntax closely follows that of [Databricks](https://docs.databricks.com/en/sql/language-manual/delta-convert-to-delta.html).

- The present implementation supports only in-place conversion, meaning that the Parquet table should be stored in a UUID dir inside the object store root (meaning bucket + any prefix).
- Note that delta-rs [doesn't support](https://github.com/delta-io/delta-rs/blob/733b5ffdde99cbfb256b8f69ae4529aeb1174599/crates/deltalake-core/src/operations/convert_to_delta.rs#L245-L248) appending/overwriting existing tables for now.
- This also lacks support for partitioned parquet tables—those could in principle be supported by extending the proposed syntax to allow passing the partitioning scheme, but since we don't yet support partitioned tables in general (see #476) I have opted out of that.
- Finally one other thing that's missing support is on-the-fly casting/coercion to delta supported arrow types subset. In other words if a Parquet file contains timestamp columns with s, ms or ns resolution this will error out.

Closes #469.

### DIFF
--- a/src/context/delta.rs
+++ b/src/context/delta.rs
@@ -3,7 +3,6 @@ use crate::context::SeafowlContext;
 #[cfg(test)]
 use crate::frontend::http::tests::deterministic_uuid;
 use crate::object_store::wrapped::InternalObjectStore;
-use crate::schema::Schema as SeafowlSchema;
 
 use bytes::BytesMut;
 use datafusion::error::{DataFusionError as Error, Result};
@@ -11,6 +10,7 @@ use datafusion::execution::context::SessionState;
 use datafusion::parquet::basic::{Compression, ZstdLevel};
 use datafusion::{
     arrow::datatypes::{Schema, SchemaRef},
+    datasource::TableProvider,
     error::DataFusionError,
     execution::context::TaskContext,
     parquet::{arrow::ArrowWriter, file::properties::WriterProperties},
@@ -18,8 +18,9 @@ use datafusion::{
     sql::TableReference,
 };
 use deltalake::kernel::{Action, Add, Schema as DeltaSchema};
-use deltalake::operations::create::CreateBuilder;
-use deltalake::operations::transaction::commit;
+use deltalake::operations::{
+    convert_to_delta::ConvertToDeltaBuilder, create::CreateBuilder, transaction::commit,
+};
 use deltalake::protocol::{DeltaOperation, SaveMode};
 use deltalake::writer::create_add;
 use deltalake::DeltaTable;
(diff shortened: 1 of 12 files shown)


src/repository/default.rs
```diff
@@ -239,10 +239,16 @@ impl Repository for $repo {
...
-        if !schema.arrow_schema.fields().is_empty() {
+        if !schema.fields().is_empty() {
             let mut builder: QueryBuilder<_> =
                 QueryBuilder::new("INSERT INTO table_column(table_version_id, name, type) ");
-            builder.push_values(schema.to_column_names_types(), |mut b, col| {
+
+            let fields: Vec<(String, String)> = schema.fields()
+                .iter()
+                .map(|f| (f.name().clone(), field_to_json(f).to_string()))
+                .collect();
```
mildbyte, line 249: Wait, it's already implemented in https://github.com/splitgraph/seafowl/blob/3d428c98bd8b83237a3cb6cae1bb48080a3c4026/src/schema.rs#L37-L43, why the duplication? could pull that out into a function that `schema::Schema` and this both call if you want to not use the `schema::Schema` wrapper struct.

src/context.rs
```diff
@@ -708,27 +709,53 @@ impl DefaultSeafowlContext {
...
+
+                // Now convert them to a Delta table
+                ConvertToDeltaBuilder::new()
+                    .with_log_store(table_log_store)
+                    .with_table_name(&*table_name)
+                    .with_comment(format!(
+                        "Converted by Seafowl {}",
+                        env!("CARGO_PKG_VERSION")
+                    ))
+                    .await?
```
mildbyte, line 744: I think we do want to support completely-zero copy, but that means that either the source files have to be in a UUID-named directory inside of a prefix, or we have to track the full path for each table in the catalog (some intersection here with being able to persist external tables, including external Delta tables, probably an extension of https://github.com/splitgraph/seafowl/pull/472 where we assume all tables have the same prefix), or we do a destructive convert-to-delta and _move_ all Parquet files to a UUID-named directory.

mildbyte (commented):
(just some initial thoughts, not a review)

src/context.rs
gruuya: > we have to track the full path for each table in the catalog

Yeah, this makes the most sense to me going forward, but we'll need to think about the new catalog schema for that.

There's also a possibility of optionally outsourcing this info (completely or partially) to a 3rd party data catalog service, e.g. see here for how delta-rs envisions that: https://github.com/delta-io/delta-rs/blob/main/crates/deltalake-core/src/data_catalog/glue/mod.rs

For now though, I'm going to go with zero-copy, and assume the files are stored at exactly the right location (bucket + optional prefix + new table UUID).

tests/statements/convert.rs
```diff
@@ -0,0 +1,75 @@
...
+    let table_uuid = Uuid::new_v4();
+    let temp_dir = maybe_test_dir.expect("temporary data dir exists");
+    let table_path = temp_dir.path().join(table_uuid.to_string());
+    // Create the directory as otherwise the COPY will fail
+    create_dir(table_path.clone()).await?;
+
+    // COPY some values multiple times to test converting flat table with more than one parquet file
+    context
+        .plan_query(&format!(
+            "COPY (VALUES (1, 'one'), (2, 'two')) TO '{}/file_1.parquet'",
```
mildbyte, line 17: I didn't realize we / DataFusion could do that 😅

mildbyte (approved):

tests/statements/convert.rs
gruuya: Yup, ever since https://github.com/splitgraph/seafowl/pull/462

src/context/delta.rs
```diff
@@ -272,20 +273,22 @@ pub async fn plan_to_object_store(
         .collect()
 }
 
+pub(super) enum CreateDeltaTableDetails {
+    WithSchema(Schema),
```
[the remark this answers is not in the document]
gruuya: Makes sense, I'm not too happy with the enum/variants name anyway.

PR #478, opened 2023-12-07
Add Arrow Flight SQL frontend

### LINKED ISSUE #477: Add Arrow Flight frontend
Currently Seafowl provides two interfaces, the HTTP frontend and the PG endpoint. While they're fine for summarized (i.e. aggregated) data output for end users directly or via a web app, they fall short when it comes to transferring large result sets. 

Putting aside the fact that not all Arrow types are supported on the currently existing frontends (e.g. see #393), returning large data sets will most likely be highly inefficient due to the overhead of converting the internal columnar representation into row-based JSON response (for the HTTP frontend) for non-trivial row counts. The particular scenario there would involve an external DB system which uses Seafowl for analytical workloads, but can't push down the entire query in some cases so it must fetch the underlying data to perform the original query itself.

Introducing the Arrow Flight (Arrow Fligt SQL in particular[1]) frontend here would solve this problem, since it provides a protocol for sending Arrow data, and would thus avoid unnecessary serialization.

[1] https://voltrondata.com/resources/apache-arrow-flight-sql-arrow-for-every-database-developer

gruuya:
Closes #477.

This PR introduces an optional Arrow Flight SQL interface to Seafowl.

1. Only ad-hoc queries supported for now
2. No prepared statements, no transactions, no write statements
3. No authentication


tests/flight/client.rs
gruuya, line 1: I'm thinking that a better test here would be to have a randomly generated batch/table with more field types, and more rows.

gruuya: Also, we'd ideally want JDBC/ODBC integration tests as well.

mildbyte: Yeah, or reuse some of the datasets we have in the HTTP tests? IIRC there's one that tests JSON serialization for everything. Probably want to test some gnarlier datatypes like Parquet structs here too.

tests/flight/client.rs
```diff
@@ -0,0 +1,65 @@
...
+        .expect("Channel connected");
+
+    let mut client = FlightClient::new(channel);
+
+    // Test the handshake works
+    let _ = client.handshake("test").await.expect("error handshaking");
+
+    // Try out the actual query
+    let cmd = CommandStatementQuery {
+        query: "SELECT * FROM flight_table".to_string(),
```
mildbyte, line 25: I'd test some basic DDL here (CREATE TABLE / CREATE EXTERNAL TABLE maybe) and a slightly more complex SELECT. Maybe also test two queries getting executed at the same time, with interleaving? i.e. start Q1, start Q2, get results for Q2, get results for Q1.

mildbyte (approved):

src/frontend/flight/sql.rs
```diff
@@ -0,0 +1,124 @@
...
+    // Fetch the result batch stream, convert to flight stream and return.
+    async fn do_get_statement(
+        &self,
+        ticket: TicketStatementQuery,
+        _request: Request<Ticket>,
+    ) -> Result<Response<<Self as FlightService>::DoGetStream>, Status> {
+        let query_id =
+            String::from_utf8_lossy(ticket.statement_handle.as_ref()).to_string();
+        let batch_stream = self
+            .fetch_stream(query_id)
```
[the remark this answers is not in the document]
gruuya: So yeah, the stream is lazily executed as soon as the Flight client on the other side starts pulling on it.

PR #479, opened 2023-12-19
Revamp catalog interface

gruuya:
The metastore used so far has been very tightly coupled to the underlying catalog implementation (i.e. a Postgres/SQLite repository), such that the lines were very blurry and the effort in adding additional catalog implementations is non-trivial. 

This PR introduces a first pass that separates concerns between the metastore (a logical abstract interface), and the actual physical persistence mechanism that stores and manages the metadata:

- The metastore API is pivoted to more closely align with a simple high-level interface (such as offered in Unity Catalog or AWS Glue).
- Another change this induces is to switch away from using db/schema/table ids in that interface (which is an implementational detail of the underlying repository).
- It also means that the terminology of the interface has been aligned with Seafowl (i.e. DataFusion), namely Catalog/Schema as opposed to Database/Collection.
- The present repository implementation that backs it up is kept mostly as is behind the interface (alongside with the terminology since that reflects the present naming of our metadata tables).


src/catalog.rs
```diff
@@ -223,310 +393,271 @@ pub trait TableCatalog: Sync + Send {
...
-                RepositoryError::SqlxError(e) => Error::SqlxError(e),
+                e => e.into(),
             })
     }
 
-    async fn delete_old_table_versions(&self, table_id: TableId) -> Result<u64, Error> {
+    async fn get(&self, name: &str) -> Result<DatabaseRecord> {
+        match self.repository.get_database(name).await {
+            Ok(database) => Ok(database),
+            Err(RepositoryError::SqlxError(sqlx::error::Error::RowNotFound)) => {
```
mildbyte, line 446: minor: could map_err here?

src/catalog.rs
```diff
@@ -223,310 +393,271 @@ pub trait TableCatalog: Sync + Send {
...
     }
 
-    async fn move_table(
+    async fn update(
         &self,
-        table_id: TableId,
+        old_catalog_name: &str,
+        old_schema_name: &str,
+        old_table_name: &str,
+        _new_catalog_name: &str, // For now we don't support moving across catalogs
```
mildbyte, line 638: change this to an assert that fails if the names aren't the same in case someone downstream tries to use this?

mildbyte (approved):

src/catalog.rs
gruuya: Yup that looks better (changed it it in `SchemaStore::get` and `TableStore::get` as well).

PR #486, opened 2024-01-04
Upgrade to DataFusion 34 and Arrow 49

gruuya:
Also bump the other deps (moka, wasm, etc.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -569,6 +569,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "backoff"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b62ddb9cb1ec0a098ad4bbf9344d0713fa193ae1a80af55febcff2627b6a00c1"
+dependencies = [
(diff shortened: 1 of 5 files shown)


src/object_store/cache.rs
```diff
@@ -576,7 +572,8 @@ mod tests {
 
         // Ensure that our TTL parameter is honored, so that we don't get a frozen cache after it
         // gets filled up once.
-        let _ = tokio::time::sleep(Duration::from_secs(1)).await;
+        let _ = tokio::time::sleep(Duration::from_millis(1100)).await;
```
mildbyte, line 575: 1.1s and not 1s? (funny I was complaining about `sleep` in tests in the previous PR and had it in my code to begin with)

mildbyte (approved):

src/object_store/cache.rs
gruuya: Oh, the previous `sleep` was put there by me (also 1s now leads to non-consistent results, sometimes a couple of entries get vacated and sometimes none at all)

PR #489, opened 2024-01-16
Implement a crude replacement for SET search_path in Flight SQL frontend

### LINKED ISSUE #487: search_path support
Support `SET search_path` either in SQL as a temporary context-altering statement:

```
-- table bla.table_1 exists
SET search_path = 'bla'; SELECT * FROM table_1;   -- when this is a multi-statement single query, e.g single HTTP / gRPC request
```

or as some metadata header in Arrow gRPC, whichever is simplest.

gruuya:
Closes #487 

Instead of adding a support for statement, enable equivalent functionality via tonic metadata values (gRPC request headers).

### DIFF
--- a/src/context/logical.rs
+++ b/src/context/logical.rs
@@ -18,8 +18,9 @@ use datafusion::optimizer::analyzer::Analyzer;
 use datafusion::optimizer::optimizer::Optimizer;
 use datafusion::optimizer::simplify_expressions::SimplifyExpressions;
 use datafusion::optimizer::{OptimizerContext, OptimizerRule};
+use datafusion::prelude::SessionContext;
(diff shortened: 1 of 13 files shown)


src/context/logical.rs
```diff
@@ -348,7 +351,7 @@ impl SeafowlContext {
 
             let full_table_name = table.to_string();
             let mut resolved_ref = TableReference::from(full_table_name.as_str())
-                .resolve(&self.database, DEFAULT_SCHEMA);
+                .resolve(&self.default_catalog, &self.default_schema);
```
gruuya, line 354: This is the one place where I don't call the new `resolve_table_ref` as I wasn't able to resolve the lifetime conundrums.

mildbyte (approved):

PR #493, opened 2024-01-24
Replace `log` with `tracing`

### LINKED ISSUE #492: Introduce JSON logging option
This would enable instrumenting Seafowl, and facilitate proper diagnostics in k8s deployments.

gruuya:
Also enable JSON logging, closes #492.

Couple of follow-ups:
- ~The default logger swallows up the target from libraries that are using plain `log` when converting them to tracing events, which obfuscates the source of those messages. They're still available in the JSON output in the `fields.log.target` field. Opened https://github.com/tokio-rs/tracing/issues/2862 for that.~
- The timestamp and target fields are dimmed now, whereas they were bold previously.


src/main.rs
```diff
@@ -167,15 +190,7 @@ async fn main() {
     }
 
     if !args.cli {
-        let mut builder = pretty_env_logger::formatted_timed_builder();
-
-        builder
-            .parse_filters(
-                env::var(env_logger::DEFAULT_FILTER_ENV)
-                    .unwrap_or_else(|_| "sqlx=warn,info".to_string())
```
mildbyte, line 175: think there was a reason for this sqlx default (it's too spammy by default or vice versa, don't remember)

mildbyte (approved):

src/main.rs
gruuya: Yeah seems like it was to spammy: 398b9947b06878d67df1bb38272db740225fd698

However I'm not seeing this anymore (nothing appears by default on INFO level).

PR #494, opened 2024-01-26
Add Linux Arm64 builds

### LINKED ISSUE #491: feat: arm64 builds
Hi, is there any change you could release arm64 binaries/docker images?
Thanks!

gruuya:
Closes #491

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -16,15 +16,18 @@ jobs:
 
     strategy:
       matrix:
-        build: [linux, macos, win-msvc]
+        build: [linux-x86_64, linux-aarch_64, osx-x86_64, win64]
         include:
-          - build: linux
+          - build: linux-x86_64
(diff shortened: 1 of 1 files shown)


onpaws:
I don't have a lot of context, apologies if we already decided this isn't possible. 
A single GHA runner seems appealing for simplicity, so assuming that single runner has to stay x86 based, I wonder if running the ARM binary inside [qemu](https://wiki.qemu.org/Documentation/Platforms/ARM) could help us do a basic 'did it run' test 🤔

gruuya:
> I wonder if running the ARM binary inside [qemu](https://wiki.qemu.org/Documentation/Platforms/ARM) could help us do a basic 'did it run' test

That's an interesting suggestion, thanks. Nonetheless, it seems simpler to wait for GH to release the linux arm64 image so that we can use that directly (otherwise I'm not sure how complicated/portable the build would get).

PR #499, opened 2024-02-08
Implement prometheus metrics

gruuya:
Current limitations:
- only metrics for the HTTP frontend
- no authN

### DIFF
--- a/src/frontend/http.rs
+++ b/src/frontend/http.rs
@@ -501,21 +501,19 @@ pub fn filters(
         .max_age(CORS_MAXAGE);
 
     let log = warp::log::custom(|info: Info<'_>| {
-        let duration = format!("{:?}", info.elapsed());
-
         #[cfg(feature = "metrics")]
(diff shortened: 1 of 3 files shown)


src/frontend/http.rs
```diff
@@ -496,7 +500,32 @@ pub fn filters(
...
+    let log = warp::log::custom(|info: Info<'_>| {
+        let duration = format!("{:?}", info.elapsed());
+
+        #[cfg(feature = "metrics")]
+        counter!(
+            HTTP_REQUESTS,
+            "method" => info.method().as_str().to_string(),
+            "path" => info.path().to_string(),
+            "status" => info.status().as_u16().to_string(),
+            //"duration" => duration.clone(),
```
gruuya, line 512: I think it's ok to include this, just a note that it will result in separate metric counts (the metrics are grouped by all fields)

mildbyte (approved):

PR #500, opened 2024-02-09
Add gRPC metrics

gruuya:
Add basic gRPC metrics (path + status), via a middleware layer intercepting request/response.

### DIFF
--- a/src/frontend/flight/metrics.rs
+++ b/src/frontend/flight/metrics.rs
@@ -21,7 +21,8 @@ impl<S> Layer<S> for MetricsLayer {
     }
 }
 
-// This service implements the Log behavior
+// This service wraps our Arrow Flight (SQL) endpoints
+// for the sole purpose of tracking metrics
 #[derive(Clone)]
 pub struct MetricsService<S> {
     inner: S,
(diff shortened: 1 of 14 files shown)


src/frontend/flight/metrics.rs
```diff
@@ -0,0 +1,96 @@
...
+                    }
+                }
+                Err(_) => {
+                    // The above means that this is some internal protocol level error or similar.
+                    // TODO: investigate when this happens.
+                    info!(
+                        path,
+                        "Error in metrics middleware while executing inner gRPC call"
+                    );
+                    Code::Unknown
```
gruuya, line 82: This code can also result by using `Status::from_error` in some of `map_err`s.

Could just use -1 in this case instead.

mildbyte: Maybe also a 500? Since this is effectively a server error

mildbyte: 2 or 13 I guess, since it's gRPC https://grpc.github.io/grpc/core/md_doc_statuscodes.html

gruuya: Ok leaving it as Unknown (2) for now, we can change it later once we realize what are the circumstances this can occur in.

PR #502, opened 2024-02-20
Add a http health endpoint

gruuya:
In particular this one corresponds to k8s readiness probe. For now it only includes trying to connect to the Flight gRPC endpoint if configured, and if that fails returns 503.


src/frontend/http.rs
```diff
@@ -596,9 +621,19 @@ pub fn filters(
         .then(upload)
         .map(into_response);
 
+    // Health-check/readiness probe
+    let ctx = context;
+    let health_route = warp::path!("readyz")
```
gruuya, line 626: Still not sure whether this is readyz, healthz or both.

src/config/context.rs
```diff
@@ -236,12 +236,12 @@ pub async fn build_context(cfg: &schema::SeafowlConfig) -> Result<SeafowlContext
     // (it will reload its schema before running the query)
 
     Ok(SeafowlContext {
+        config: Arc::new(cfg),
```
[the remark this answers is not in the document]
mildbyte: Yeah, might as well make it an owned struct without the Arc

src/frontend/http.rs
```diff
@@ -482,8 +484,29 @@ async fn load_part(mut part: Part) -> Result<Vec<u8>, ApiError> {
     Ok(bytes)
 }
 
-// We need the allow to silence the compiler: it asks us to add warp::generic::Tuple to the first
-// parameter of the return type, but that struct is not exportable (generic is private).
+/// GET /readyz
+pub async fn health_endpoint(context: Arc<SeafowlContext>) -> Result<Response, ApiError> {
+    #[cfg(feature = "frontend-arrow-flight")]
+    if let Some(flight) = &context.config.frontend.flight {
+        // TODO: run SELECT 1 or something similar?
```
mildbyte, line 491: Depends, does us being able to serve this endpoint mean the Arrow Flight frontend is initialized already? Does the Flight frontend check for connectivity on startup by loading the catalog?

src/frontend/http.rs
mildbyte: IIUC healthz (liveness) is for when something internal has gone wrong in the system and we need it to be restarted. E.g. we check that some background worker thread is still running and if it's crashed, we say we're unhealthy.

I think we can make `healthz` and `readyz` do the same thing for now.

src/frontend/http.rs
```diff
@@ -596,9 +621,19 @@ pub fn filters(
...
+        .and(warp::path::end())
+        .and(warp::get())
+        .and(warp::any().map(move || ctx.clone()))
+        .then(health_endpoint)
+        .map(into_response);
+
     cached_read_query_route
         .or(uncached_read_write_query_route)
         .or(upload_route)
+        .or(health_route)
```
mildbyte, line 636: Maybe this should be a separate frontend, since the user can have a gRPC frontend enabled without HTTP. But then gRPC isn't authed and HTTP is, so it doesn't make it less secure.

mildbyte (approved):

src/config/context.rs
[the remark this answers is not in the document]
gruuya: I now remembered that I initially put the Arc on top of it due to the `scope_to_*` methods which create a new context on the heap, but it's probably ok to do it anyway since the config is tiny.

src/frontend/http.rs
gruuya: Yeah, prefer to leave it there for now, but if there's a strong need can separate it out later on.

src/frontend/http.rs
gruuya: > does us being able to serve this endpoint mean the Arrow Flight frontend is initialized already?

In principle no, since there's a short period during startup when Flight frontend might not have initialized but the HTTP one has. In practice probably yes, since Flight frontend is spawned first, and there is likely a miniscule startup time for both.

> Does the Flight frontend check for connectivity on startup by loading the catalog?

No, just create a server listening fore queries. Only when one arrives does it try to load the catalog and resolve the identifiers.

PR #508, opened 2024-03-11
Add caching on top of clade-provided object stores too

gruuya:


src/catalog/metastore.rs
```diff
@@ -147,11 +159,19 @@ impl Metastore {
...
+                let mut store = stores
+                    .get(&location)
+                    .ok_or(CatalogError::Generic {
+                        reason: format!("Object store for location {location} not found"),
+                    })?
+                    .clone();
+
+                if let Some(ref cache) = self.object_store_cache {
+                    store = cache.wrap_store(store);
+                }
```
gruuya, line 171: Should skip this in case of local FS, however unlikely that may be.

src/catalog/metastore.rs
```diff
@@ -70,9 +73,18 @@ impl Metastore {
             functions: external_store,
             staging_schema,
             default_store,
+            object_store_cache: None,
         }
     }
 
+    pub fn with_object_store_cache(
+        mut self,
```
mildbyte, line 81: (a bit more idiomatic)

```suggestion
        &mut self,
```

mildbyte (approved):

src/catalog/metastore.rs
gruuya: I guess it's down to preference whether the builder consumes/returns a thing, or updates the reference of the thing: https://users.rust-lang.org/t/builder-pattern-in-rust-self-vs-mut-self-and-method-vs-associated-function/72892/4

I was mostly guided by the object_store crate which goes with the former approach.

PR #512, opened 2024-04-02
Extend assume role tests

gruuya:
Add an example using OIDC via a AssumeRoleWithWebIdentity action, with a mock OIDC provider backed by Dex.


docker-compose.yml
```diff
@@ -53,3 +68,62 @@ services:
     entrypoint: >
       curl -v -X POST --data-binary '{"name":"seafowl-test-bucket"}' -H "Content-Type:
       application/json" "http://fake-gcs:4443/storage/v1/b"
+
+configs:
+  dex.yaml:
```
gruuya, line 73: It's nice that this works, but since the config is not small it might make sense to actually extract it into a proper file.

tests/main.rs
```diff
@@ -145,30 +146,81 @@ async fn get_addr() -> SocketAddr {
...
+            // Exchange Dex token for valid MinIO STS credentials using the AssumeRoleWithWebIdentity
+            // action.
+            let config = aws_config::SdkConfig::builder()
+                .region(aws_config::Region::new("us-east-1"))
+                .endpoint_url("http://localhost:9000")
+                .build();
+
+            let creds = Client::new(&config)
+                .assume_role_with_web_identity()
+                .web_identity_token(dex_token.as_str().unwrap())
```
gruuya, line 212: This is a low-level API in-line with the above case. I could also try to use the higher level API and actually persist/load the token from the file: https://docs.rs/aws-config/latest/aws_config/web_identity_token/index.html#

PR #515, opened 2024-04-16
Best effort chunk coalescing implementation

gruuya:
This is a naive implementation of https://github.com/splitgraph/seafowl/pull/514 (and therefore less complex), where we don't use synchronization to avoid request de-duplucation, but instead try do a best-effort chunk coalescing, and batch them into a single call.

Compared to current main, this implementation seems to be more stable, and doesn't experience the hanging in TPC-H q18 and q21, since it does not use any synchronization primitives.

As for perf, in the (idealized) no latency case it is on par with the current caching object store. However in more realistic scenarios of modest delays (15ms) it consistently outperforms it:
image

### DIFF
--- a/src/object_store/cache.rs
+++ b/src/object_store/cache.rs
@@ -18,7 +18,6 @@ use std::fmt::{Debug, Formatter};
 
 use std::fs::remove_dir_all;
 use std::io;
-use std::io::ErrorKind;
 use std::ops::Range;
 use std::path::{Path, PathBuf};
 
@@ -45,12 +44,10 @@ impl CacheFileManager {
         let mut path = self.base_path.to_path_buf();
         path.push(cache_key.as_filename());
 
-        // Should this happen normally?
+        // TODO: when does this happen?
         if path.exists() {
-            return Err(io::Error::new(
-                ErrorKind::Other,
-                "Internal error: cached file path already exists",
-            ));
+            debug!("{cache_key:?} file already exists, skipping write");
+            return Ok(path.clone());
         }
 
         tokio::fs::write(&path, data.as_ref()).await?;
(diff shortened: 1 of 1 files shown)


src/object_store/cache.rs
```diff
@@ -41,20 +40,14 @@ impl CacheFileManager {
...
-        &self,
-        cache_key: &CacheKey,
-        data: Arc<Bytes>,
-    ) -> io::Result<PathBuf> {
+    async fn write_file(&self, cache_key: &CacheKey, data: Bytes) -> io::Result<PathBuf> {
         let mut path = self.base_path.to_path_buf();
         path.push(cache_key.as_filename());
 
-        // Should this happen normally?
+        // TODO: when does this happen?
```
mildbyte, line 47: I guess when:

thread 1 evicts the item from cache, notifies the eviction listener (but that doesn't delete the file yet)
thread 2 tries to get the item, cache miss, downloads it, tries to write to the file

I wish there were some kind of moka semantics where the cache entry stays there as long as and only as long as the eviction listener hasn't finished :D

src/object_store/cache.rs
```diff
@@ -217,88 +215,145 @@ impl CachingObjectStore {
...
+                                Err(err) => {
+                                    warn!(
+                                        "Re-downloading cache value for {key:?}: {err}"
+                                    );
+                                    let data = self
+                                        .inner
+                                        .get_range(location, chunk_range.clone())
+                                        .await?;
+                                    self.cache_chunk_data(key, data.clone()).await;
+                                    Some(data)
```
mildbyte, line 280: Worth adding a comment here to clarify the race condition this mitigates against:

thread 1 gets a value from cache, it's File(some-path) (self.cache.get on line 254)
thread 2 runs eviction, deletes the value and the file
thread 1 tries to read some-path, it's gone now

mildbyte (approved):

src/object_store/cache.rs
mildbyte: It's a shame there isn't some get-with Moka functionality that takes in a closure that the value from getting evicted while it's running (dual to the comment above)

src/object_store/cache.rs
gruuya: Yeah, agreed with both comments.

PR #516, opened 2024-04-17
Move Delta store creation into a separate ObjectStoreFactory

mildbyte:
The factory makes sure that object_store clients are only built once (using the URL and options as the key, no expiry) instead of on every call. This means that the cached data from the same object store will be shared across several queries and won't get reset.

### DIFF
--- a/src/object_store/factory.rs
+++ b/src/object_store/factory.rs
@@ -1,0 +1,218 @@
+/// Contains logic for constructing and caching object stores for various
+/// purposes, including wrapping them in a bunch of layers for Delta Lake,
+/// HTTP object stores, caching etc
+use std::{
+    collections::HashMap,
+    hash::Hash,
+    sync::{Arc, RwLock},
+};
+
+use deltalake::{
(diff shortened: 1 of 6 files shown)


src/object_store/factory.rs
```diff
@@ -0,0 +1,218 @@
...
+        let mut pairs: Vec<_> = self.options.iter().collect();
+        pairs.sort_by_key(|i| i.0);
+
+        Hash::hash(&pairs, state);
+    }
+}
+
+pub struct ObjectStoreFactory {
+    default_store: Arc<InternalObjectStore>,
+    custom_stores: RwLock<HashMap<StoreCacheKey, Arc<dyn ObjectStore>>>,
```
gruuya, line 120: Let's consider doing this, though we can do it in a follow-up PR.
```suggestion
    custom_stores: DashMap<StoreCacheKey, Arc<dyn ObjectStore>>>,
```

gruuya (approved):

PR #517, opened 2024-04-18
Improve Arrow Flight SQL interface logging

gruuya:
Log query ids, request metadata as well as errors.

### DIFF
--- a/src/frontend/flight/handler.rs
+++ b/src/frontend/flight/handler.rs
@@ -11,7 +11,7 @@ use std::sync::Arc;
 use tokio::sync::Mutex;
 use tonic::metadata::MetadataMap;
 use tonic::Status;
-use tracing::debug;
+use tracing::{error, info};
 
 lazy_static! {
     pub static ref SEAFOWL_SQL_DATA: SqlInfoData = {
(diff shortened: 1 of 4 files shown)


src/frontend/flight/handler.rs
```diff
@@ -61,8 +62,14 @@ impl SeafowlFlightHandler {
...
-        let plan = ctx.plan_query(query).await?;
-        let batch_stream = ctx.execute_stream(plan).await?;
+        let plan = ctx
+            .plan_query(query)
+            .await
+            .inspect_err(|err| debug!("Error planning query id {query_id}: {err}"))?;
+        let batch_stream = ctx
+            .execute_stream(plan)
+            .await
+            .inspect_err(|err| debug!("Error executing query id {query_id}: {err}"))?;
```
mildbyte, line 72: These could be INFO I think (since we can't turn on debug logging on the fly without redeploying Seafowl)

src/frontend/flight/handler.rs
```diff
@@ -73,14 +80,12 @@ impl SeafowlFlightHandler {
...
+        query_id: &str,
     ) -> core::result::Result<SendableRecordBatchStream, Status> {
-        let (_, batch_stream_mutex) =
-            self.results
-                .remove(&query_id)
-                .ok_or(Status::not_found(format!(
-                    "No results found for query id {query_id}"
-                )))?;
+        let (_, batch_stream_mutex) = self.results.remove(query_id).ok_or_else(|| {
+            debug!("No results found for query id {query_id}");
```
mildbyte, line 86: Should be an INFO/ERROR? If a client is requesting a query that it thought we were executing but we aren't, we definitely want to know

mildbyte (approved):

PR #520, opened 2024-04-29
DataFusion 37.1 upgrade

gruuya:
Also bump Arrow to v51, and patch to pick up https://github.com/apache/datafusion/pull/10114,

### DIFF
--- a/tests/clade/query.rs
+++ b/tests/clade/query.rs
@@ -1,7 +1,7 @@
 use crate::clade::*;
 
 #[rstest]
-#[should_panic(expected = "table 'default.local.file' not found")]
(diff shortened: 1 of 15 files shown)


tests/clade/query.rs
```diff
@@ -1,7 +1,7 @@
 use crate::clade::*;
 
 #[rstest]
-#[should_panic(expected = "table 'default.local.file' not found")]
+#[should_panic(expected = "External(NotATable(\"no log files\"))")]
```
mildbyte, line 4: oh this is useful, I got confused so many times by this showing up when looking at invalid table paths

mildbyte (approved):

PR #521, opened 2024-04-29
Remove obsolete key from docker-compose.yml

SergeiPatiakin:
## What
Remove obsolete 'version' key from from docker-compose.yml
Our CI configuration uses Docker Compose v2, which ignores the 'version' key.

## Why
Reduce noise when running docker compose

## Context
https://docs.docker.com/compose/intro/history/


mildbyte (approved):
thanks!

PR #525, opened 2024-05-14
Ignore DeltaTableAlready delta error during conversion of a parquet folder

gruuya:


src/context/delta.rs
```diff
@@ -333,15 +333,32 @@ impl SeafowlContext {
...
                     .with_save_mode(SaveMode::Overwrite)
                     .with_comment(format!(
                         "Converted by Seafowl {}",
                         env!("CARGO_PKG_VERSION")
                     ))
-                    .await?;
+                    .await
+                {
+                    Ok(table) => table,
+                    // Match on table already exist error
```
mildbyte, line 348: (to avoid some WTFs)

```suggestion
                    // Match on table already exist error
                    // We can't match on the enum value directly, since this
                    // Error enum isn't public
```

mildbyte (approved):

PR #527, opened 2024-05-17
Seafowl data replication initial implementation

gruuya:


src/frontend/flight/handler.rs
```diff
@@ -89,4 +104,121 @@ impl SeafowlFlightHandler {
...
+            return Ok(DataSyncResult {
+                accepted: true,
+                memory_sequence_number: mem_seq,
+                durable_sequence_number: dur_seq,
+            });
+        }
+
+        debug!("Processing data change with {num_rows} rows for url {url}");
+        // TODO: make timeout configurable
+        let accepted =
```
mildbyte, line 190: Come to think about it, we might also need to reject the write if it'll put us above some "hard" memory limit. Maybe we should have a "low watermark" (flush when we're using more RAM than this) and "high watermark" (reject writes to never go above this limit) memory limit. We'll find out after load testing I guess.

src/frontend/flight/sync.rs
```diff
@@ -0,0 +1,293 @@
...
+            syncs: Default::default(),
+            size: 0,
+        }
+    }
+
+    // Extract the latest memory sequence number for a given table location.
+    pub fn mem_seq_for_table(&self, url: &String) -> Option<u64> {
+        self.syncs
+            .get(url)
+            .map(|entry| entry.syncs.last().unwrap().sequence_number)
```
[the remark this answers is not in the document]
gruuya: Can't use flatmap I think (it's not an iterable that iIm collecting), but I can go with
```diff
     pub fn mem_seq_for_table(&self, url: &String) -> Option<u64> {
         self.syncs
             .get(url)
-            .map(|entry| entry.syncs.last().unwrap().sequence_number)
+            .map(|entry| entry.syncs.last().map(|s| s.sequence_number))
+            .flatten()
     }
```
which avoids the use of `unwrap`.

mildbyte (approved):

src/frontend/flight/sync.rs
```diff
@@ -0,0 +1,293 @@
...
+        {
+            // TODO: do this out-of-band
+            Some(self.lags.pop().unwrap().0)
+        } else if self.size >= self.context.config.misc.sync_data.max_in_memory_bytes {
+            // TODO: maybe better to pop from a size-based max-heap in this case
+            Some(self.lags.pop().unwrap().0)
+        } else {
+            None
+        }
+    }
```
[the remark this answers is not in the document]
gruuya: > will this first flush the entire first table and then flush the entire second table?

That is correct. My thinking was that we'd prefer to have this done in one go, and thus avoid fragmentation.

Also my impression was that in the general case two tables won't be having matching sequence numbers anyway?

tests/flight/sync.rs
```diff
@@ -0,0 +1,277 @@
...
+    client: &mut FlightClient,
+) -> Result<DataSyncResult> {
+    let flight_data = sync_cmd_to_flight_data(cmd.clone(), batch);
+    let response = client.do_put(flight_data).await?.next().await.unwrap()?;
+
+    Ok(DataSyncResult::decode(response.app_metadata).expect("DataSyncResult"))
+}
+
+#[tokio::test]
+async fn test_sync_happy_path() -> std::result::Result<(), Box<dyn std::error::Error>> {
```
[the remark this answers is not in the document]
gruuya: It doesn't test that but it does test whether the empty payload returns the correct current memory/durable state:
https://github.com/splitgraph/seafowl/blob/5354567a2c918a57e882b93e086856e2b4be5ed6/tests/flight/sync.rs#L205-L215

Agreed though, there's a bunch of unit tests for the `SeafowlSyncManager` that should be added.

src/frontend/flight/handler.rs
gruuya: Maybe? I guess if we're over the low watermark we're going to be flushing syncs into storage and reducing the memory usage anyway, it's not like we're going to continue accumulating stuff in memory.

src/frontend/flight/handler.rs
```diff
@@ -89,4 +104,121 @@ impl SeafowlFlightHandler {
...
+            .iter()
+            .fold(0, |rows, batch| rows + batch.num_rows());
+        if num_rows == 0 {
+            debug!("Received empty batches, returning current sequence numbers");
+            return Ok(DataSyncResult {
+                accepted: true,
+                memory_sequence_number: mem_seq,
+                durable_sequence_number: dur_seq,
+            });
+        }
```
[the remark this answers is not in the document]
gruuya: Ah something like "if this hasn't been invoked during the lifetime of `SeafowlSyncManager` return this in all cases and set some flag, otherwise return only if the flag was set **and** num_rows == 0"?

src/frontend/flight/handler.rs
```diff
@@ -89,4 +104,121 @@ impl SeafowlFlightHandler {
...
+            let commit_infos = table.history(Some(1)).await?;
+            match commit_infos
+                .last()
+                .expect("Table has non-zero commits")
+                .info
+                .get(SEAFOWL_SYNC_DATA_SEQUENCE_NUMBER)
+            {
+                Some(Value::Number(seq)) => seq.as_u64(),
+                _ => None,
+            }
```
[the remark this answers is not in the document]
gruuya: > we'd have schema version numbers

You mean SeafowlCommitMetadata JSON schema version right?

Yeah, that could make sense, though I'd be hesitant to do it in this PR.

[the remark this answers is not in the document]
mildbyte: Right, not in this PR, just something to keep in mind - as soon as we start persisting metadata in Delta Tables, we'll have to think about the schema and schema migrations if we add/remove metadata fields.

src/frontend/flight/handler.rs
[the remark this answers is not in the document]
mildbyte: When a new client first connects (basically, we want to make sure that if Seafowl restarts and the client reconnects, the client knows which offset to restart from, so it sends an empty DoPut without any data to get the startup LSNs, and, as a precaution, Seafowl would reject DoPuts that aren't empty).

src/frontend/flight/handler.rs
mildbyte: True, if the flushing happens in the same process as the Flight handler, if it happens in the background, we might keep accumulating data even though we started the flush? Guess we'll find out with load testing

src/frontend/flight/sync.rs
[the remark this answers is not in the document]
mildbyte: Unless this unwrap is really an assertion (at this point in time, if the URL exists, there must be at least one entry for it)

src/frontend/flight/sync.rs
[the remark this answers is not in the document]
mildbyte: As discussed offline, the sequence numbers are for the whole change stream, not individual tables, so we'll need to rearrange this a bit.

tests/flight/sync.rs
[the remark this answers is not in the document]
mildbyte: Yup, this test looks good and agreed on deeper "logic" tests in the sync code 👍

clade/src/lib.rs
```diff
@@ -3,3 +3,30 @@ pub mod schema {
     pub const FILE_DESCRIPTOR_SET: &[u8] =
         tonic::include_file_descriptor_set!("clade_descriptor");
 }
+
+pub mod sync {
+    use arrow_flight::sql::{Any, ProstMessageExt};
+    use prost::Message;
+
+    tonic::include_proto!("clade.sync");
```
eatonphil, line 11: Why are we doing this and not just letting the proto be in plain Rust?

gruuya: Fair question—I think a part of the answer is that we're using flight as the actual interface here, and it is protobuf based, so this is too by extension.

On a more practical side we may want to enable clients in different languages using this, and this would facilitate that.

src/frontend/flight/sync.rs
```diff
@@ -0,0 +1,378 @@
...
+        } else if self.size >= self.context.config.misc.sync_data.max_in_memory_bytes {
+            // Or if we're over the size limit flush the oldest entry
+            Some(self.syncs.first().unwrap().0.clone())
+        } else {
+            None
+        }
+    }
+
+    async fn flush_batches(&mut self) -> Result<()> {
+        while let Some(url) = self.flush_ready() {
```
mildbyte, line 199: The fact that we flush (and commit LSNs to Delta) url-by-url still breaks my brain - I need to think more about this tomorrow.

Say I have:

```
insert(table_1, [item_1])
insert(table_2, [item_2])
modify(table_2, [add item_3])
insert(table_3, [item_4]))
modify(table_2, [add item_5])
modify(table_3, [add item_6])
```

(and each item is a single-batch transaction, i.e. last=true always)

does the queue look like this?

```
[highest priority item, first to flush]
1. table_1: item_1
2. table_2: item_2, item_3, item_5
3. table_3: item_4, item_6
```

so if we flush only the first two items, will we say that we're durable up to 5, which is not true because we haven't flushed item_4 from table_3 yet?

src/frontend/flight/sync.rs
```diff
@@ -0,0 +1,378 @@
...
+        if let Some((url, sync)) = self.syncs.first()
+            && now() - sync.insertion_time
+                >= self.context.config.misc.sync_data.max_replication_lag_s
+        {
+            // First flush any changes that are past the configured max duration
+            // TODO: do this out-of-band
+            Some(url.clone())
+        } else if self.size >= self.context.config.misc.sync_data.max_in_memory_bytes {
+            // Or if we're over the size limit flush the oldest entry
+            Some(self.syncs.first().unwrap().0.clone())
```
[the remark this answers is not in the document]
gruuya: That is correct, but it's not a strict limitation, we can also do that.

src/frontend/flight/sync.rs
```diff
@@ -0,0 +1,378 @@
...
+                    Ok((col(name, schema.as_ref())?, name.to_string()))
+                } else {
+                    Ok((
+                        lit(ScalarValue::Null.cast_to(f.data_type())?),
+                        name.to_string(),
+                    ))
+                }
+            })
+            .collect::<Result<_>>()?;
+        let proj_plan = Arc::new(ProjectionExec::try_new(projection, filter_plan)?);
```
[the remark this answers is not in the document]
gruuya: Well the arrow writer complains prior to it so delta may be ok with it. This is bound to change with the proposed follow-up work anyway though.

src/frontend/flight/sync.rs
gruuya: You're right about the state of the queue, and that this is a potential problem, will think about ways to resolve.

src/frontend/flight/sync.rs
```diff
@@ -376,3 +432,109 @@ fn now() -> u64 {
...
+            .unwrap();
+
+        assert_eq!(sync_mgr.stored_sequences(&origin), (Some(3), Some(1)),);
+
+        // Wait for the replication lag to exceed the configured max duration
+        tokio::time::sleep(Duration::from_secs(1)).await;
+
+        // We've now advanced durable past seq 2 and on to seq 3
+        sync_mgr.flush_batches().await.unwrap();
+        assert_eq!(sync_mgr.stored_sequences(&origin), (Some(3), Some(3)),);
```
mildbyte, line 538: Does table_2 w/LSN2 get flushed at this point too (it should, since we advanced the LSN to 3)?

src/frontend/flight/sync.rs
gruuya: One potential resolution is outlined in 666a70dadc84f38f6733e68b10a5cf33049c2636. It's not simple but it's a first pass and can be simplified further if we decide to go along with this approach.

src/frontend/flight/sync.rs
```diff
@@ -368,6 +381,49 @@ impl SeafowlDataSyncManager {
...
+
+                let mut remove_count = 0;
+                for (seq_num, seq) in origin_seqs.into_iter() {
+                    if seq.locs.is_empty() && seq.last {
+                        // We've seen the last sync for this sequence, all pending locations
+                        // have been flushed to and there's no preceding sequence to be flushed,
+                        // so we're good to flag the sequence as durable
+                        self.origin_durable.insert(origin.clone(), *seq_num);
+
+                        remove_count += 1;
```
mildbyte, line 407: Hm, in this example:

```
insert (table_1, lsn_1, last=true)
insert (table_2, lsn_2, last=true)
insert (table_1, lsn_3, last=true)
```

When we flush only table_1, what's preventing us from looking at `lsn_3`, seeing that it has no tables pending to flush (is that correct? do we add table_2 to this list for lsn_3 as well?) and advancing the durable LSN=3?

src/frontend/flight/sync.rs
```diff
@@ -376,3 +432,109 @@ fn now() -> u64 {
...
+                vec!["c1".to_string()],
+                true,
+                vec![dummy_batch()],
+            )
+            .await
+            .unwrap();
+
+        assert_eq!(sync_mgr.stored_sequences(&origin), (Some(2), None),);
+
+        // Add LSN 3 for table 1 and push it over the max configured size
```
mildbyte, line 518: What would happen if you added a change for table_2 instead (this would also trigger a flush) - would we flush just the changes for table_1 and advance the LSN to 1?

src/frontend/flight/sync.rs
```diff
@@ -368,6 +381,49 @@ impl SeafowlDataSyncManager {
 
         Ok(Arc::new(UnionExec::new(vec![input_plan, proj_plan])))
     }
+
+    // Based on the provided origins and sequence numbers deduce what is the current durable
+    // sequence number for each origin.
+    fn advance_durable(
```
mildbyte, line 387: Highly recommend adding some unit tests for this function, since this is where the most important logic lives and it has lots of mutability of seqs/locs/origin_durable

src/frontend/flight/sync.rs
gruuya: Yup, the only reason we can advance to LSN 3 is because table_2 was flushed as well, will add some asserts to make it explicit.

src/frontend/flight/sync.rs
gruuya: Because we've exited the for loop for lsn 2 (in particular `seq.locs.is_empty()` is false in that case since the table 2 location is still pending).

src/frontend/flight/sync.rs
gruuya: I've now made a parametrizable test called `test_sync_flush` and added some cases.

Let me know what you think, and especially let me know about any cases I can add there.

src/frontend/flight/sync.rs
```diff
@@ -0,0 +1,620 @@
...
+// Each `DataSyncCommand` that accompanies a record batch carries information on
+// the origin of change, sequence number, primary keys and whether this is the
+// last message in the sequence.
+//
+// It uses a greedy table-based algorithm for flushing: once the criteria is met
+// it will go through table's ordered by the oldest sync flushing all pending syncs
+// for that table. Special care is taken about deducing what is the correct durable
+// sequence to report back to the caller.
+//
+//           │sync #1│sync #2│sync #3│sync #4│sync #5│
```
eatonphil, line 44: sync here refers to the result of an RPC? a batch basically?

src/frontend/flight/sync.rs
```diff
@@ -0,0 +1,620 @@
...
+//
+// Finally, once table_3 is flushed `SeafowlDataSyncManager` will advance the
+// durable sequence up to 3, since both it and 2 have now been completely persisted.
+pub(super) struct SeafowlDataSyncManager {
+    context: Arc<SeafowlContext>,
+    // All sequences kept in memory, queued up by insertion order, per origin,
+    seqs: HashMap<String, IndexMap<SequenceNumber, DataSyncSequence>>,
+    // An indexed queue of table URL => pending batches to upsert/delete
+    // sorted by insertion order
+    syncs: IndexMap<String, DataSyncCollection>,
```
eatonphil, line 76: What is the difference between these? seqs and syncs?

src/frontend/flight/sync.rs
```diff
@@ -0,0 +1,620 @@
...
+    // Total size of all batches in memory currently
```
eatonphil, line 77: The (what feels like) inconsistent use of the term "batch" (versus "sync" or "seq"?) makes it hard for me to understand some of the comments

src/frontend/flight/sync.rs
```diff
@@ -0,0 +1,620 @@
...
+    size: usize,
+    // Map of known memory sequence numbers per origin
+    origin_memory: HashMap<String, SequenceNumber>,
+    // Map of known durable sequence numbers per origin
+    origin_durable: HashMap<String, SequenceNumber>,
```
eatonphil, line 82: Why are we treating Origin as a String and not a u64?

clade/proto/sync.proto
```diff
@@ -0,0 +1,43 @@
...
+
+  // Primary key columns. Must not be empty: the table must always
+  // have a replica identity / PK.
+  repeated string pk_column = 3;
+
+  // Opaque string identifying data source
+  string origin = 4;
+
+  // Monotonically-increasing transaction number (e.g. the LSN)
+  uint64 sequence_number = 5;
```
eatonphil, line 24: Just note that it can and does wrap.

clade/proto/sync.proto
```diff
@@ -0,0 +1,43 @@
...
+
+  // True if this is the last command in the transaction denoted
+  // by this LSN and Seafowl can advance its LSN after flushing
```
eatonphil, line 27: It would help if we consistently used LSN or Seq as a term.

tests/flight/sync.rs
```diff
@@ -0,0 +1,280 @@
...
+    flight_stream_builder.build(futures::stream::iter(vec![Ok(batch)]))
+}
+
+async fn do_put_sync(
+    cmd: DataSyncCommand,
+    batch: RecordBatch,
+    client: &mut FlightClient,
+) -> Result<DataSyncResult> {
+    let flight_data = sync_cmd_to_flight_data(cmd.clone(), batch);
+    let response = client.do_put(flight_data).await?.next().await.unwrap()?;
```
eatonphil, line 19: Why do we do await().next().await?

clade/proto/sync.proto
```diff
@@ -0,0 +1,43 @@
...
+}
+
+message DataSyncResult {
+  // If false, Seafowl is overloaded and can't accept the change (e.g.
+  // can't flush fast enough). The client should wait and retry.
+  // TODO: also schema mismatch errors
+  bool accepted = 1;
+
+  // Sequence number up to which the changes are in Seafowl's memory.
+  optional uint64 memory_sequence_number = 2;
```
eatonphil, line 39: Why do you return this?

src/frontend/flight/sync.rs
gruuya: Yes, sync corresponds to a single RPC call (so a vector of batches with the same schema).

src/frontend/flight/sync.rs
gruuya: `seqs` is a map (of maps) of sequences that are still held in memory. We track the insertion time and whether the sequence has had its last sync call yet (and is thus a potential durable candidate), and also keep them ordered for each origin.

`syncs` is a map of table locations with pending flushes.

src/frontend/flight/sync.rs
gruuya: That's a good question, my impression was that these would be strings.

src/frontend/flight/sync.rs
gruuya: The bottom term actually refers to (the size of) all the batches in memory.

The upper one is more descriptive, it's really a map of table URL => syncs (which in turn keep store the actual batches plus some additional info). Let me update that one.

clade/proto/sync.proto
gruuya: Thanks, that's a very important point. How often does this occur roughly?

clade/proto/sync.proto
gruuya: Ok, dropping LSN in favor of sequence throughout.

clade/proto/sync.proto
gruuya: If you're referring to the fact that it is optional, this is so because after a restart we have nothing in memory, and it feels inadequate to convey this through some actual number (e.g. 0 or -1).

Otherwise if you mean why return this at all, I think it's useful for when e.g. the client (sending) side restarts and wants to replay something—it can then skip some stuff as per this value.

tests/flight/sync.rs
gruuya: Because the `do_put` call is itself async (first await), but it also returns a [stream](https://github.com/apache/arrow-rs/blob/c2b05cdbcb37f46f170c7b5073ee6bc2178fdace/arrow-flight/src/sql/server.rs#L382) which we need to await in order to fetch the result. 

In our case we only stream one element back (result) so it''s not very useful, but the mechanism is there for more complex stuff.

src/frontend/flight/handler.rs
```diff
@@ -89,4 +103,96 @@ impl SeafowlFlightHandler {
...
+            return Ok(DataSyncResult {
+                accepted: true,
+                memory_sequence_number: mem_seq,
+                durable_sequence_number: dur_seq,
+            });
+        }
+
+        debug!("Processing data change with {num_rows} rows for url {url}");
+        // TODO: make timeout configurable
+        match tokio::time::timeout(Duration::from_secs(3), self.sync_manager.write())
```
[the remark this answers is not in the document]
gruuya: It's configurable now :D

clade/proto/sync.proto
PJMODOS: LSN does not wrap

mildbyte: They shouldn't wrap as per https://www.postgresql.org/docs/16/wal-internals.html and other discussions:

> The numbers do not wrap, but it will take a very, very long time to exhaust the available stock of numbers.

clade/proto/sync.proto
mildbyte: Will also be useful later on as per the discussion in 

https://docs.google.com/document/d/1v1HCaw-__fx8JM5DwRe23lc4pt_w8XAC-wRkFoegrbc/edit?disco=AAABKFZw9wA

> the system in general needs to know what's in memory to be able to provide reasonable snapshots for queries

src/frontend/flight/sync.rs
```diff
@@ -0,0 +1,620 @@
...
+        vec![None, Some(1), Some(4)]
+    )]
+    #[case::long_sequence_mid_flush(
+        &[("table_1", 1), ("table_1", 1), ("table_1", 1), FLUSH, ("table_1", 1), ("table_2", 1),
+            ("table_2", 2), ("table_2", 2), FLUSH, ("table_2", 2), ("table_3", 2), FLUSH, ("table_3", 3),
+            FLUSH, ("table_3", 3), ("table_1", 4), ("table_3", 4), ("table_1", 4), FLUSH, ("table_3", 4),
+            FLUSH, FLUSH],
+        // Reasoning for the observed durable sequences:
+        // - seq 1 not seen last sync
+        // - seq 1 seen last sync, but it is in a unflushed table (2)
```
mildbyte, line 554: I got confused as to why this is None and not 1, but it's because after the previous flush, we received data for table_1 and table_2, and the first flush after that only flushes table_1?

src/frontend/flight/sync.rs
```diff
@@ -0,0 +1,620 @@
...
+            Field::new(SEAFOWL_SYNC_DATA_UD_FLAG, DataType::Boolean, false),
+        ]));
+
+        // Generate a random length between 1 and 3
+        let len: usize = rand::thread_rng().gen_range(1..=3);
+
+        (0..len)
+            .map(|_| create_random_batch(schema.clone(), 10, 0.2, 0.8).unwrap())
+            .collect()
+    }
```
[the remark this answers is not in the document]
gruuya: No, the values are way below the default size-bound flush threshold.

src/frontend/flight/sync.rs
gruuya: That is correct, each `FLUSH` in the test executes a single flush call, which translates to a single table location with the oldest pending changes.

src/context/delta.rs
```diff
@@ -188,10 +184,11 @@ pub async fn plan_to_object_store(
...
 
                 // For local FS stores, we can just move the file to the target location
-                if let Some(result) =
-                    store.fast_upload(&partition_file_path, &location).await
-                {
-                    result?;
+                if let Some(data_dir) = local_data_dir {
+                    // NB: in order to exploit the fast upload path for local FS store we need to
+                    // provide the absolute path to the destination files
+                    let location = format!("{data_dir}/{file_name}");
```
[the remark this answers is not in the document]
mildbyte: eh no need, as discussed

src/frontend/flight/sync.rs
mildbyte: Looking good! Suggested a couple cases.

clade/proto/sync.proto
eatonphil: My mistake, apologies.

clade/proto/sync.proto
```diff
@@ -0,0 +1,43 @@
...
+
+  // Root URL (e.g. s3://my-bucket) and connection options for the table's path
+  schema.StorageLocation store = 2;
+
+  // Primary key columns. Must not be empty: the table must always
+  // have a replica identity / PK.
+  repeated string pk_column = 3;
+
+  // Opaque string identifying data source
+  string origin = 4;
```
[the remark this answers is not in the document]
gruuya: Fixed now, thanks.

clade/proto/sync.proto
```diff
@@ -0,0 +1,43 @@
...
+  // Will be created based on the schema of the RecordBatch if doesn't exist.
+  // e.g. path/to/table
+  string path = 1;
+
+  // Root URL (e.g. s3://my-bucket) and connection options for the table's path
+  schema.StorageLocation store = 2;
+
+  // Primary key columns. Must not be empty: the table must always
+  // have a replica identity / PK.
+  repeated string pk_column = 3;
```
eatonphil, line 18: Should this be called `pk_columns`? (plural)

gruuya: Good point, it is indeed a [recommendation](https://protobuf.dev/programming-guides/style/#repeated-fields), revising.

mildbyte (approved):

PR #529, opened 2024-06-07
Add support for deletes and updates that don't touch primary keys

gruuya:
This PR introduces:
- qualifier generation from record batches
- delta table partition file pruning
- applying sync data to build an input plan to stream through

### DIFF
--- a/src/frontend/flight/sync.rs
+++ b/src/frontend/flight/sync.rs
@@ -8,7 +8,9 @@ use datafusion::prelude::DataFrame;
 use datafusion_common::{
     Column, DataFusionError, JoinType, Result, ScalarValue, ToDFSchema,
 };
+use datafusion_expr::aggregate_function::AggregateFunction as AggregateFunctionFunc;
 use datafusion_expr::execution_props::ExecutionProps;
+use datafusion_expr::expr::AggregateFunction;
 use datafusion_expr::{
     col, is_false, is_null, lit, when, LogicalPlanBuilder, UNNAMED_TABLE,
 };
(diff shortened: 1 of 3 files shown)


src/frontend/flight/sync.rs
```diff
@@ -371,42 +400,170 @@ impl SeafowlDataSyncManager {
         )
     }
 
-    fn plan_data_sync(
+    // Get a list of files that will be scanned and removed based on the pruning qualifier.
+    fn get_removes(
```
mildbyte, line 404: Basically, all these partitions are the ones that we're going to overwrite, so we'll delete these ones and then add our merged ones?

src/frontend/flight/sync.rs
```diff
@@ -266,89 +271,113 @@ impl SeafowlDataSyncManager {
...
+            &input_plan,
+            log_store.object_store(),
+            local_data_dir,
+            self.context.config.misc.max_partition_size,
+        )
+        .await?;
+
+        let mut actions: Vec<Action> = adds.into_iter().map(Action::Add).collect();
+
+        // Prune away files that are refuted by the qualifier
```
mildbyte, line 349: as in partitions that we overwrote, right?

src/frontend/flight/sync.rs
```diff
@@ -371,42 +400,170 @@ impl SeafowlDataSyncManager {
...
-        // TODO: Filter away deletes for now
-        let filter_plan: Arc<dyn ExecutionPlan> = Arc::new(FilterExec::try_new(
-            col(SEAFOWL_SYNC_DATA_UD_FLAG, schema.as_ref())?,
-            mem_plan,
-        )?);
+        let sync_df = self.context.inner.read_batches(data)?;
+
+        // First remove all deleted rows
+        let join_cols = pk_columns.iter().map(|pk| pk.as_str()).collect::<Vec<_>>();
+        let input_df = input_df.join(
```
mildbyte, line 515: What if a row is updated multiple times in a single batch (or updated then deleted, or deleted then updated)? Will we take the latest version in that case? (I think this might need a window function with `row_number() OVER (PARTITION BY pk ORDER BY seq_number ASC`)

src/frontend/flight/sync.rs
```diff
@@ -371,42 +400,170 @@ impl SeafowlDataSyncManager {
...
+            .join(
+                sync_df.filter(col(SEAFOWL_SYNC_DATA_UD_FLAG))?,
+                JoinType::Full,
+                &join_cols,
+                &join_cols,
+                None,
+            )?
+            .select(projection)?;
+
+        Ok(input_df)
```
mildbyte, line 566: Will this write an empty partition if the entire partition gets cleaned out by a sync?

src/frontend/flight/sync.rs
gruuya: That is correct, the same as when planning a DELETE/UPDATE.

src/frontend/flight/sync.rs
gruuya: Yup, correct.

src/frontend/flight/sync.rs
gruuya: Good point, haven't tried that out (for some reason i assumed this won't happen), need to think about it.

src/frontend/flight/sync.rs
gruuya: It shouldn't write an empty partition but it may write an empty commit (without adds/removes). 

We could do some pre-exec aggregation later on to e.g. see how many rows would be produced, and if it's zero and there were no removes just skip the commit?

tests/flight/sync.rs
```diff
@@ -126,40 +149,47 @@ async fn test_sync_happy_path() -> std::result::Result<(), Box<dyn std::error::E
...
-                1149573966000000,
+                None,
+                None,
+                Some(1115269505000000),
+                Some(1149573966000000),
+            ])),
+            Arc::new(StringArray::from(vec![
+                Some("five"),
+                Some("five"),
+                Some("five"),
```
mildbyte, line 188: The second one should be None (since we're deleting it) and I suggest the final one should be different so that we can be confident that the last value won? Or does the fact that the timestamp goes from None -> 1115269505000000 test this?

```suggestion
                None,
                Some("five_updated_after_delete"),
```

src/frontend/flight/sync.rs
mildbyte: Oh yeah I see, since this goes into `plan_to_object_store`. Let's test for what happens when an entire write is just deletes (later, after we've settled on the protocol).

mildbyte (approved):

tests/flight/sync.rs
gruuya: > The second one should be None (since we're deleting it)

Can't do that now since I've made c2 be a part of PK as well.

> Or does the fact that the timestamp goes from None -> 1115269505000000 test this?

Exactly, only in the last row do we get an actual timestamp (which is shown in the query results).

PR #530, opened 2024-06-13
Sync protocol improvements

gruuya:
The changes in this PR involve:
- new more robust replication protocol (formalized in the clade proto files).
- batch compaction, meaning compressing changes from a single sync into a minimal form getting rid of temporary intermediate changes
- idempotent CRUD support

TODOS:
- Further segmentation of logic pieces that are not strictly needed in the sync writer
- A lot more unit testing, and a bit more integration testing

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ arrow-flight = "51.0.0"
 # https://github.com/apache/arrow-rs/pull/2868
 # https://github.com/apache/arrow-rs/pull/2724
 arrow-integration-test = "51.0.0"
+arrow-row = "51.0.0"
 arrow-schema = "51.0.0"
 async-trait = "0.1.64"
 
@@ -71,6 +72,7 @@ arrow-flight = { workspace = true, optional = true }
 # https://github.com/apache/arrow-rs/pull/2868
 # https://github.com/apache/arrow-rs/pull/2724
 arrow-integration-test = { workspace = true }
+arrow-row = { workspace = true }
 arrow-schema = { workspace = true }
 async-trait = { workspace = true }
 base64 = "0.21.0"
--- a/clade/proto/sync.proto
+++ b/clade/proto/sync.proto
@@ -4,6 +4,22 @@ package clade.sync;
 
 import "schema.proto";
 
+enum ColumnRole {
+  // Denotes a column that that has the old value for a primary key column
+  OLD_PK = 0;
+  // Denotes a column that that has the new value for a primary key column
+  NEW_PK = 1;
+  // Denotes whether a particular column has changed in value
+  CHANGED = 2;
+  // Denotes the value of a non-PK column
+  VALUE = 3;
+}
+
+message ColumnDescriptor {
+  ColumnRole role = 1;
+  string name = 2;
+}
+
 message DataSyncCommand {
   // Path to the Delta Table, relative to the root URL of the store below.
   // Will be created based on the schema of the RecordBatch if doesn't exist.
@@ -13,9 +29,8 @@ message DataSyncCommand {
   // Root URL (e.g. s3://my-bucket) and connection options for the table's path
   schema.StorageLocation store = 2;
 
-  // Primary key columns. Must not be empty: the table must always
-  // have a replica identity / PK.
-  repeated string pk_columns = 3;
+  // Metadata about the columns in the RecordBatch
+  repeated ColumnDescriptor column_descriptors = 3;
 
   // Integer identifying data source
   uint64 origin = 4;
--- a/src/frontend/flight/sync/schema.rs
+++ b/src/frontend/flight/sync/schema.rs
@@ -1,0 +1,129 @@
+use crate::frontend::flight::sync::SyncError;
+use arrow_schema::{DataType, FieldRef, SchemaRef};
+use clade::sync::{ColumnDescriptor, ColumnRole};
+use tracing::warn;
+
+#[derive(Debug, Clone)]
+pub struct SyncSchema {
+    column_descriptors: Vec<ColumnDescriptor>,
+    schema: SchemaRef,
+}
+
+impl SyncSchema {
+    pub fn try_new(
+        column_descriptors: Vec<ColumnDescriptor>,
+        schema: SchemaRef,
+    ) -> Result<Self, SyncError> {
+        if column_descriptors.len() != schema.all_fields().len() {
+            let err = format!(
+                "Column descriptors {:?} do not match the schema {schema}",
+                column_descriptors
+            );
+            return Err(SyncError::SchemaError { reason: err });
+        }
+
+        // Validate field role's are parsable, we have the correct number of old/new PKs,
+        // and Changed role is non-nullable boolean type which points to an existing column
+        // TODO: Validate types on old and new PKS
+        let mut old_pk_count = 0;
+        let mut new_pk_count = 0;
+        for (col_desc, field) in column_descriptors.iter().zip(schema.fields()) {
+            match ColumnRole::try_from(col_desc.role) {
+                Ok(ColumnRole::OldPk) => old_pk_count += 1,
+                Ok(ColumnRole::NewPk) => new_pk_count += 1,
+                Ok(ColumnRole::Value) => {}
+                Ok(ColumnRole::Changed) => {
+                    let err = if field.data_type() != &DataType::Boolean {
+                        format!(
+                            "Field for column with `Changed` role must be of type boolean: {}",
+                            field.name()
+                        )
+                    } else if field.is_nullable() {
+                        format!(
+                            "Field for column with `Changed` role can not be nullable: {}",
+                            field.name()
+                        )
+                    } else if !column_descriptors.iter().any(|other_cd| {
+                        col_desc.name == other_cd.name
+                            && other_cd.role == ColumnRole::Value as i32
+                    }) {
+                        format!(
+                            "Column with `Changed` role must point to an existing column with `Value` role: {}",
+                            &col_desc.name
+                        )
+                    } else {
+                        // All good
+                        continue;
+                    };
+                    warn!(err);
+                    return Err(SyncError::SchemaError { reason: err });
+                }
+                Err(err) => {
+                    let err = format!("Failed parsing role from field metadata: {err:?}");
+                    warn!(err);
+                    return Err(SyncError::SchemaError { reason: err });
+                }
+            }
+        }
+        if old_pk_count == 0 || new_pk_count == 0 || old_pk_count != new_pk_count {
+            let err = "Change requested but batches do not contain old/new PK columns";
+            warn!(err);
+            return Err(SyncError::SchemaError {
+                reason: err.to_string(),
+            });
+        }
+
+        Ok(Self {
+            column_descriptors,
+            schema,
+        })
+    }
+
+    #[allow(dead_code)]
+    pub fn arrow_schema(&self) -> SchemaRef {
+        self.schema.clone()
+    }
+
+    pub fn column(&self, name: &str, role: ColumnRole) -> Option<SyncColumn> {
+        self.columns()
+            .find(|col| col.name == name && col.role == role)
+    }
+
+    pub fn columns(&self) -> impl Iterator<Item = SyncColumn> + '_ {
+        self.column_descriptors
+            .iter()
+            .zip(self.schema.fields())
+            .map(|(column_descriptor, field)| SyncColumn {
+                role: column_descriptor.role(),
+                name: column_descriptor.name.clone(),
+                field: field.clone(),
+            })
+    }
+}
+
+pub struct SyncColumn {
+    role: ColumnRole,
+    name: String,
+    field: FieldRef,
+}
+
+impl SyncColumn {
+    // Get the role of the column
+    pub fn role(&self) -> ColumnRole {
+        self.role
+    }
+
+    // Name of the column that this descriptor refers to. Note that in general this differs from the
+    // underlying arrow field name. You can think of them as physical and logical names, respectively.
+    //
+    // For all roles except `Changed`, the sync column refers to itself, but for `Changed` it refers
+    // to a `Value` column whose field name is the same as this name.
+    pub fn name(&self) -> &String {
+        &self.name
+    }
+
+    // Get the field from the arrow schema
+    pub fn field(&self) -> &FieldRef {
+        &self.field
+    }
+}
--- a/src/frontend/flight/sync/schema.rs
+++ b/src/frontend/flight/sync/schema.rs
@@ -25,6 +25,7 @@ impl SyncSchema {
         // Validate field role's are parsable, we have the correct number of old/new PKs,
         // and Changed role is non-nullable boolean type which points to an existing column
         // TODO: Validate types on old and new PKS
+        // TODO: Validate a column can not be a PK and Value at the same time
         let mut old_pk_count = 0;
         let mut new_pk_count = 0;
         for (col_desc, field) in column_descriptors.iter().zip(schema.fields()) {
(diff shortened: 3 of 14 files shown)


clade/proto/sync.proto
```diff
@@ -4,6 +4,22 @@ package clade.sync;
 
 import "schema.proto";
 
+enum ColumnRole {
+  // Denotes a column that that has the old value for a primary key column
+  OLD_PK = 0;
+  // Denotes a column that that has the new value for a primary key column
+  NEW_PK = 1;
+  // Denotes whether a particular column has changed in value
+  CHANGED = 2;
```
eatonphil, line 13: I am still not personally confident about how TOAST columns actually work. We may need to change more here or not. But nbd for now, just mentioning.

src/frontend/flight/sync/schema.rs
```diff
@@ -0,0 +1,128 @@
...
+        self.role
+    }
+
+    // Name of the column that this descriptor refers to. Note that in general this differs from the
+    // underlying arrow field name. You can think of them as physical and logical names, respectively.
+    //
+    // For all roles except `Changed`, the sync column refers to itself, but for `Changed` it refers
+    // to a `Value` column whose field name is the same as this name.
+    pub fn name(&self) -> &String {
+        &self.name
```
eatonphil, line 121: How are you allowed to return a reference without a lifetime?

src/frontend/flight/sync/utils.rs
```diff
@@ -0,0 +1,425 @@
...
+        converter.convert_columns(&nulls)?,
+    ))
+}
+
+// Compact a set of record batches into a single one, squashing any chain of changes to a given row
+// into a single row in the output batch.
+// This means that if a row is changed multiple times, only the last change will be reflected in the
+// output batch (meaning the last NewPk and Value role columns and the last Value column where the
+// accompanying Changed field was `true`).
+pub(super) fn compact_batches(
```
eatonphil, line 41: While I see the point about squashing changes, isn't it pretty wasteful to be converting to rows and then back to columns?

eatonphil: Also, can't you squash rows without actually converting to row format?

src/frontend/flight/sync/utils.rs
```diff
@@ -0,0 +1,425 @@
...
+                )?)
+            })
+            .collect::<Result<_>>()?,
+    )?;
+
+    Ok(batch)
+}
+
+// Generates a pruning qualifier for the table based on the PK columns and the min/max values
+// in the data sync entries.
```
eatonphil, line 203: I don't quite understand the point of this, like what is the benefit? The comment doesn't help me much

src/frontend/flight/sync/schema.rs
gruuya: It's an implicit reference, I think the compiler inferred it from `&self`

eatonphil: Shows how little I know :)

src/frontend/flight/sync/utils.rs
gruuya: Fair point; for one thing note this was designed to be efficiently encodable: 
https://github.com/apache/arrow-rs/blob/8752e01be642bce205984e16b44e06078413dc68/arrow-row/src/lib.rs#L20-L24

In fact DataFusion itself relies on it for some core computations itself (also we only ever transpose the PKs, though that isn't a big factor anyway).

gruuya: > Or I'm not sure what you mean

Ah, so what needs to happen above is the squashing (compaction) of all chains of changes to a given row (that may encompass multiple input rows), into a single output row, which will then be applied to the target table.

These chains can be of arbitrary lengths, from 1 (meaning a row is touched only once) to the batch row count (meaning the entire batch conveys changes for a single original row). In addition, various things can happen in each chain, such as value changes, pk changes and entire row deletion.

In principle I think this kind of computation can be formalized using recursive CTEs. However there are 3 things to note here:

1. I briefly experimented with recursive CTEs, managing to get some basic (but not full) functionality that we need here. The simplest prototype worked ok, however slightly more involved stuff made DF hang and spike the CPU, meaning there are some bugs there and I can't resolve those now.
2. Even if there were no issues in the execution, the recursive CTE implementation (using dataframe/plan/expression primitives) would be much more complex, since it would approach the problem in a round-about way.
3. Ultimately, even such an implementation would entail using the `arrow-row` under the hood somewhere (e.g. in the grouping/window aggregation phases).

src/frontend/flight/sync/utils.rs
gruuya: The goal is to reduce the amount of data needed to be scanned/streamed/joined with from the base table.

Basically we only need to touch those partition which might have the extreme (min/max) PK values that are contained in the sync data.

src/frontend/flight/sync/utils.rs
mildbyte: Yeah, I think normally we could do this with something like this:

```
WITH latest_rows AS (
  SELECT row_number() OVER (PARTITION BY pk ORDER BY timestamp DESC) AS _row_num
  FROM table
) SELECT [cols] FROM latest_rows WHERE _row_num = 1 AND NOT is_deleted
```

which would still process things row-by-row in the background but at least would reduce the amount of our custom code that we have to write.

But with PK-changing UPDATEs we have to be clever. It does raise the question of why we expect the API user to convert row-based changes into Arrow and then this code transposes it back into rows, then back again into columnar (maybe V2 of this would be Avro changes instead).

We can always see how this performs under benchmarks (append-only, update-intensive, update-intensive with PK changes). Ultimately, we'll have to do this transpose/compact somewhere - on write or on every read.

Cargo.toml
```diff
@@ -10,6 +10,7 @@ arrow-flight = "51.0.0"
 # https://github.com/apache/arrow-rs/pull/2868
 # https://github.com/apache/arrow-rs/pull/2724
 arrow-integration-test = "51.0.0"
+arrow-row = "51.0.0"
```
mildbyte, line 13: say "arrow-row" three times out loud 🙃

src/frontend/flight/sync/schema.rs
```diff
@@ -0,0 +1,152 @@
...
+            })
+            .collect();
+
+        Ok(Self { schema, columns })
+    }
+
+    #[allow(dead_code)]
+    pub fn arrow_schema(&self) -> SchemaRef {
+        self.schema.clone()
+    }
```
mildbyte, line 100: Why keep this code if it's dead, some future impl?

src/frontend/flight/sync/schema.rs
```diff
@@ -0,0 +1,152 @@
...
+                            field.name()
+                        )
+                    } else if field.is_nullable() {
+                        format!(
+                            "Field for column with `Changed` role can not be nullable: {}",
+                            field.name()
+                        )
+                    } else if !column_descriptors.iter().any(|other_cd| {
+                        col_desc.name == other_cd.name
+                            && other_cd.role == ColumnRole::Value as i32
```
mildbyte, line 50: Is this going to be robust to us adding new members / changing the grpc enum? I guess we need to make sure that we upgrade the client / server at the same time or only add new enum members at the end?

src/frontend/flight/sync/utils.rs
```diff
@@ -0,0 +1,545 @@
...
+    let nulls = null_rows.row(0);
+
+    let (changed_binding, _) = columns(ColumnRole::Changed);
+    let changed = changed_binding
+        .iter()
+        .map(|array| {
+            array
+                .as_any()
+                .downcast_ref::<arrow::array::BooleanArray>()
+                .expect("Changed column must be boolean")
```
mildbyte, line 75: Aside: do you know what happens if we panic in the Flight handler? Does the entire Seafowl process die or it's just that the gRPC caller gets an error response?

src/frontend/flight/sync/writer.rs
```diff
@@ -193,15 +183,20 @@ impl SeafowlDataSyncManager {
             })
             .or_insert(IndexMap::from([(sequence_number, sequence)]));
 
-        // Finally upsert the new sync item for this location
-        let size = batches
-            .iter()
-            .fold(0, |bytes, batch| bytes + batch.get_array_memory_size());
+        let batch = compact_batches(&sync_schema, batches)?;
```
mildbyte, line 186: Maybe we should do this compaction once on flush instead of on every write. Otherwise in a chain of writes we'd be loading the (constantly growing) staged in-memory batch into a hashmap, then transposing back into rows on every write? Although in that case we won't support running queries with unflushed in-memory data.

src/frontend/flight/sync/utils.rs
gruuya: Got it, changing the doc.

> Aside: if my entry has two changes to PK=1 and PK=infinity, will this make me read the entire Delta table?

Yup, I don't think there's a way around it.

src/frontend/flight/sync/utils.rs
```diff
@@ -0,0 +1,545 @@
...
+            ColumnDescriptor {
+                role: ColumnRole::Value as i32,
+                name: "c4".to_string(),
+            },
+        ];
+
+        let sync_schema = SyncSchema::try_new(column_descriptors, schema.clone())?;
+
+        let mut rng = rand::thread_rng();
+        let row_count = rng.gen_range(1..=1000); // With more than 1000 rows the test becomes slow
```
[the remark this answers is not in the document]
gruuya: The test only, to be precise the random row generator part (the compaction part isn't affected).

src/frontend/flight/sync/schema.rs
gruuya: Potentially, but also I use it in the tests.

src/frontend/flight/sync/schema.rs
gruuya: There is probably a evolution of the protocol that would break this somehow, but yeah, in general adding new enum members at the end and simultaneous client / server upgrades should resolve it.

src/frontend/flight/sync/utils.rs
gruuya: Good question, I think it might be just the flight frontend task that dies? Need to check.

src/frontend/flight/sync/writer.rs
gruuya: No no, it doesn't work like that. The record batch(es) for each sync are compacted and stored individually as `DataSyncItem`s in a `DataSyncCollection` vector, meaning the previously compacted batches are never touched again until flushing. So we don't do any more work than if we do it lazily (since we'd need to do it then anyway).

The idea was to compress the size eagerly, so that we keep as low memory profile as possible.

src/frontend/flight/sync/utils.rs
```diff
@@ -0,0 +1,545 @@
...
+        ) = vec![
+            (None, Some(1), Some(1.0), Some(false), None),
+            (None, Some(2), Some(2.0), Some(true), Some("two")),
+            (Some(1), Some(1), Some(1.1), Some(true), Some("one")),
+            (Some(2), Some(-2), Some(2.1), Some(false), None),
+            (Some(3), Some(3), Some(3.3), Some(true), Some("three")),
+            (Some(4), Some(4), Some(4.0), Some(false), Some("four")),
+            (Some(1), Some(2), Some(1.2), Some(true), None),
+            (Some(3), None, None, Some(false), None),
+        ]
```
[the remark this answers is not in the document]
gruuya: Ok, I extended the test a bit (it also includes a 2-hop PK change chain now).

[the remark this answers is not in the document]
gruuya: > (as an aside, maybe we should somehow use this partition by PK / order by changed_ts / take most recent behaviour if we know there's no PK changes in the batch since that's a simpler implementation, and then use the hashmap-based approach)

I think this would only increase complexity since we'd need to take care of both scenarios then (batches w/ and w/o PK changes). Also even then we'd need special casing for the Changed/Value not being necessarily from the last row.

Furthermore, the current mechanism also in theory enables us to support actual PK evolution (type and/or count) in future if we decide we want it.

src/frontend/flight/sync/utils.rs
mildbyte: I think we can still work around it though - by manually reading the Delta metadata and cross-referencing it against each PK we know we have changes for, but that sounds like more of a future optimization to me.

src/frontend/flight/sync/schema.rs
mildbyte: You can make it non-public then (it'll still be available to unit tests in the same file)

src/frontend/flight/sync/writer.rs
mildbyte: Ah I see. What happens if there's a PK changing UPDATE in a future batch that affects a batch that we already compacted?

gruuya: Any PK changing UPDATEs in a future batch must reference the last new PK from a preceding batch (which is the one we keep), so compaction shouldn't affect it. This should be guaranteed by source database PK conflict detection mechanism.

In effect the problem is self-contained, i.e. independent of any upcoming changes.

src/frontend/flight/sync/schema.rs
eatonphil: > and simultaneous client / server upgrades should resolve it.

Always sketchy. :) We have version protocols within pgd. Perhaps we should think about that here too. (Not necessarily this second)

src/frontend/flight/sync/utils.rs
gruuya: But delta-rs already does that for us automatically, i.e. the scan uses the filter from this function to prune the partitions. So if we have PK=1..infinity (I'm assuming you mean here a really large number encompassing all partitions) then we must hit all partitions otherwise we're not guaranteeing idempotency (e.g. we may double-insert)?

src/frontend/flight/sync/schema.rs
gruuya: The unit test is in another file, so it's simpler to remove and revise the test setup slightly.

src/frontend/flight/sync/utils.rs
gruuya: Tried it, apparently nothing noteworthy happens apart from the client receiving an error:
```
Tonic(Status { code: Cancelled, message: "h2 protocol error: http2 error: stream error received: stream no longer needed", source: Some(tonic::transport::Error(Transport, hyper::Error(Http2, Error { kind: Reset(StreamId(1), CANCEL, Remote) }))) })
```
and seafowl logging
```
thread 'flight::client::test_basic_queries' panicked at src/frontend/flight/sql.rs:50:9:
Test handler panic
stack backtrace:
...
```
for each panic.

I interleaved a couple of calls to the handshake (in which I've added a panic) in some tests and while that particular grpc call has the error the rest finish successfully.

src/frontend/flight/sync/utils.rs
mildbyte: I was thinking just two changes, one hitting partition #1, one hitting the last partition, not a range of changes spanning all partitions. Surely in this case we can optimize by only rewriting these two partitions?

src/frontend/flight/sync/schema.rs
mildbyte: Right, but in this case both the server (seafowl) and the client (seafowl_pg) are part of the same distribution and will be upgraded in lockstep, as opposed to different nodes in the cluster.

src/frontend/flight/sync/utils.rs
gruuya: Ah, yeah that should be possible. 

Not sure how complicated it would be though, naively in the general case we'd need to have a qualifier per change/sync row and consequently a multiple scans/joins. 

The less naive approach would be to segment all changes and group them by partitions they hit, and then only scan/join per group.

mildbyte (approved):

PR #531, opened 2024-06-18
Add support for inline-ing metastore over Flight

gruuya:
## What

Adds support for discrete, temporary, metastores inlined with queries called over arrow flight. These can be used to resolve identifiers parsed from the query in an ad-hoc manner.

## How
- introduce a `MemoryStore` and implement all store traits for it
- encapsulate the `ListSchemaResponse` from regular clade metastore in a new message called `InlineMetastoreCommandStatementQuery` that also holds the query
- now when `get_sql_info` is called arrow fails to pattern match against known commands and uses the fallback method
- there we simply instantiate the `MemoryStore` for a one-off context that serves the query

### DIFF
--- a/src/frontend/flight/handler.rs
+++ b/src/frontend/flight/handler.rs
@@ -1,18 +1,20 @@
+use crate::catalog::memory::MemoryStore;
+use crate::catalog::metastore::Metastore;
 use arrow::record_batch::RecordBatch;
 use arrow_flight::sql::metadata::{SqlInfoData, SqlInfoDataBuilder};
-use arrow_flight::sql::SqlInfo;
-use arrow_schema::SchemaRef;
+use arrow_flight::sql::{ProstMessageExt, SqlInfo, TicketStatementQuery};
+use arrow_flight::{FlightDescriptor, FlightEndpoint, FlightInfo, Ticket};
 use clade::sync::{DataSyncCommand, DataSyncResult};
 use dashmap::DashMap;
 use datafusion::common::Result;
 use datafusion::execution::SendableRecordBatchStream;
 use datafusion_common::DataFusionError;
 use lazy_static::lazy_static;
+use prost::Message;
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::sync::{Mutex, RwLock};
-use tonic::metadata::MetadataMap;
-use tonic::Status;
+use tonic::{Request, Status};
 use tracing::{debug, error, info};
 use url::Url;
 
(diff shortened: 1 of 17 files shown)


tests/flight/mod.rs
```diff
@@ -33,19 +35,28 @@ use seafowl::frontend::flight::run_flight_server;
...
+    };
+
     let config_text = format!(
         r#"
-[object_store]
-type = "memory"
+{object_store_section}
 
 [catalog]
 type = "sqlite"
```
mildbyte, line 62: Probably the only thing left in this PR is making this catalog optional (expect the metastore to be provided inline if it isn't specified) and erroring out if the user tries to run DDL (creating/deleting tables), which we might be getting automatically because of the `not_impl()` thing

tests/flight/mod.rs
```diff
@@ -100,7 +111,26 @@ async fn get_flight_batches(
...
+async fn get_flight_batches_inlined(
+    client: &mut FlightClient,
+    query: String,
+    schemas: ListSchemaResponse,
+) -> Result<Vec<RecordBatch>> {
+    let cmd = InlineMetastoreCommandStatementQuery {
+        query,
+        schemas: Some(schemas),
+    };
+    get_flight_batches_inner(client, cmd.as_any()).await
```
mildbyte, line 126: It's nice to see that the usage example for this is very straightforward

src/frontend/flight/handler.rs
```diff
@@ -74,6 +77,14 @@ impl SeafowlFlightHandler {
             self.context.clone()
         };
 
+        if let Some(memory_store) = memory_store {
+            // If metastore was inlined with the query use it for resolving tables/locations
+            ctx = ctx.with_metastore(Arc::new(Metastore::new_from_memory(
```
mildbyte, line 82: As discussed, we might need to "merge" the in-memory and the real from-config metastore and fall back to the latter metastore if there's a table we can't find in the former, but we can do this in the separate PR if need be

mildbyte (approved):

tests/flight/mod.rs
gruuya: How about making it optional and default to in-mem sqlite? 

One can still provide the inline metastores, but we also support the full DDL spectrum, albeit ephemerally.

mildbyte: It's possible, but kind of awkward? Since if you first create a table with the inline metastore in use (e.g. use some CREATE TABLE AS statement), you wouldn't be able to query it unless you stopped providing the inline metastore, so I don't yet see how useful that would be? It's the responsibility of the caller to be the metastore/catalog if the inline metastore is being used.

gruuya: Hmm, well you can't create a table with the inline metastore at all, only reads are supported. This is because we implement only the basic `SchemaStore` for the `MemoryStore`, so we the table creation metastore calls would error out (due to `not_impl`).

mildbyte: Oh I see, you meant just defaulting to the in-memory metastore if the inline metastore isn't provided. I don't really have strong feelings about it. I guess in the Seafowl configuration where you're using it with the inline metastore it'd surface errors to the developer faster if they forgot to provide it on some code path instead of quietly using the in-mem SQLite on would be my only argument for erroring out.

PR #532, opened 2024-06-19
Upgrade to DataFusion 39

gruuya:
Also bump Arrow to v52.

### DIFF
--- a/src/datafusion/parser.rs
+++ b/src/datafusion/parser.rs
@@ -27,7 +27,6 @@
 
 pub use datafusion::sql::parser::Statement;
 use datafusion::sql::parser::{CopyToSource, CopyToStatement, CreateExternalTable};
-use datafusion_common::parsers::CompressionTypeVariant;
 use lazy_static::lazy_static;
 use sqlparser::ast::{CreateFunctionBody, Expr, ObjectName, OrderByExpr, Value};
 use sqlparser::tokenizer::{TokenWithLocation, Word};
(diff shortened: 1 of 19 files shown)


src/datafusion/parser.rs
```diff
@@ -51,6 +49,15 @@ fn parse_file_type(s: &str) -> Result<String, ParserError> {
...
+fn ensure_not_set<T>(field: &Option<T>, name: &str) -> Result<(), ParserError> {
+    if field.is_some() {
+        return Err(ParserError::ParserError(format!(
+            "{name} specified more than once",
+        )));
+    }
+    Ok(())
+}
+
 // XXX SEAFOWL: removed the struct definitions here because we want to use
```
mildbyte, line 61: I forgot we keep having to pull DF's parser changes and rebase ours on top of them, does DF still not have hooks to extend the parser?

src/nodes.rs
```diff
@@ -175,6 +175,14 @@ impl UserDefinedLogicalNode for SeafowlExtensionNode {
         Arc::from(self.clone())
     }
 
+    fn with_exprs_and_inputs(
+        &self,
+        _exprs: Vec<Expr>,
+        _inputs: Vec<LogicalPlan>,
+    ) -> datafusion_common::Result<Arc<dyn UserDefinedLogicalNode>> {
+        todo!("Should we implement this")
```
mildbyte, line 183: Yeah, looks like it's the same thing as `from_template`, so use that impl? I'm surprised we don't crash with this `todo()`.

https://github.com/apache/datafusion/blob/5bfc11ba4ac4f11eaf9793c668e4a064fb697e6e/datafusion/expr/src/logical_plan/extension.rs#L79-L103

```suggestion
        Ok(Arc::from(self.clone()))
```

mildbyte (approved):

src/datafusion/parser.rs
gruuya: Yes I meant to look into that, but didn't.

Looks like it's easier now, but still some methods are just private so we'd still need to copy most of the stuff.

PR #533, opened 2024-06-20
Add sync metrics

gruuya:
Add metrics tracking byte size/row count for:
- incoming request
- compaction
- flushing
as well as timings for the last two ops.

### DIFF
--- a/src/frontend/flight/sync/metrics.rs
+++ b/src/frontend/flight/sync/metrics.rs
@@ -1,12 +1,18 @@
-use metrics::{describe_gauge, gauge, Gauge};
+use metrics::{
+    counter, describe_counter, describe_gauge, describe_histogram, gauge, histogram,
+    Gauge, Histogram,
+};
 
 const IN_MEMORY_SIZE: &str = "seafowl_sync_writer_in_memory_size";
 const IN_MEMORY_ROWS: &str = "seafowl_sync_writer_in_memory_rows";
+const COMPACTION_TIME: &str = "seafowl_sync_writer_compaction_time";
+const COMPACTED: &str = "seafowl_sync_writer_compacted";
 
 #[derive(Clone)]
 pub struct SyncMetrics {
     pub in_memory_size: Gauge,
     pub in_memory_rows: Gauge,
+    pub compaction_time: Histogram,
 }
 
 impl Default for SyncMetrics {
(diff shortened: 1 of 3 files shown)


src/frontend/flight/sync/metrics.rs
```diff
@@ -0,0 +1,95 @@
...
+    counter, describe_counter, describe_gauge, describe_histogram, gauge, histogram,
+    Gauge, Histogram,
+};
+
+const REQUEST: &str = "seafowl_sync_writer_request";
+const IN_MEMORY: &str = "seafowl_sync_writer_in_memory";
+const COMPACTION_TIME: &str = "seafowl_sync_writer_compaction_time";
+const COMPACTED: &str = "seafowl_sync_writer_compacted";
+const FLUSHING_TIME: &str = "seafowl_sync_writer_flushing_time";
+const FLUSHED: &str = "seafowl_sync_writer_flushed";
```
mildbyte, line 11: https://prometheus.io/docs/practices/naming/

The rule of thumb for these is that if you sum up some metric by all possible label values, it should still be a meaningful value, so we can't have different units in the same metric, instead:

```
# counter: total rows requested and total size of requests
seafowl_sync_writer_requests_rows_total
seafowl_sync_writer_requests_bytes_total

# gauges: how many rows in memory and how much space do they take
seafowl_sync_writer_in_memory_rows_total
seafowl_sync_writer_in_memory_bytes_total

# histogram: time taken by compaction
seafowl_sync_writer_compaction_time_seconds

# counter: total rows compacted, amount and size
seafowl_sync_writer_compacted_rows_total
seafowl_sync_writer_compacted_bytes_total

# histogram: time taken by flushes
seafowl_sync_writer_flushed_time_seconds

# counter: total rows flushed, amount and size
seafowl_sync_writer_flushed_bytes_total
seafowl_sync_writer_flushed_rows_total
```

src/frontend/flight/sync/metrics.rs
```diff
@@ -0,0 +1,95 @@
+use metrics::{
+    counter, describe_counter, describe_gauge, describe_histogram, gauge, histogram,
+    Gauge, Histogram,
+};
+
+const REQUEST: &str = "seafowl_sync_writer_request";
```
mildbyte, line 6: I think we should rename this `sync` to something else, `seafowl_delta_writer`? `seafowl_writer`? `seafowl_changeset_writer`?

src/frontend/flight/sync/writer.rs
```diff
@@ -380,6 +410,12 @@ impl SeafowlDataSyncWriter {
         self.remove_sync(&url);
         self.advance_durable();
 
+        // Record flush metrics
+        let flush_duration = start.elapsed().as_millis();
+        self.metrics.flushing_time.record(flush_duration as f64);
+        self.metrics.flushed_size(size as u64);
+        self.metrics.flushed_rows(rows as u64);
+
         Ok(())
```
mildbyte, line 419: Let's also add a gauge here for timestamp of last flush

```
seafowl_writer_last_successful_flush_timestamp_seconds
```

src/frontend/flight/sync/writer.rs
```diff
@@ -380,6 +410,12 @@ impl SeafowlDataSyncWriter {
         self.remove_sync(&url);
         self.advance_durable();
```
mildbyte, line 411: We should also expose the LSNs here (I think they're in bytes):

```
# gauge
seafowl_sync_writer_durable_lsn_bytes{origin => 123}
seafowl_sync_writer_memory_lsn_bytes{origin => 123}
```

src/frontend/flight/sync/metrics.rs
gruuya: I'll go with `seafowl_changeset_writer`.

src/frontend/flight/sync/writer.rs
gruuya: Good idea, adding.

src/frontend/flight/sync/metrics.rs
```diff
@@ -0,0 +1,119 @@
...
+        describe_histogram!(
+            COMPACTION_TIME,
+            "The time taken to compact a single sync message"
+        );
+        describe_counter!(
+            COMPACTED_BYTES,
+            "The reduction in byte size due to batch compaction"
+        );
+        describe_counter!(
+            COMPACTED_BYTES,
```
mildbyte, line 79:
```suggestion
            COMPACTED_ROWS,
```

mildbyte (approved):

PR #534, opened 2024-06-21
Add flush task

gruuya:
Try to flush on a regular interval, so as not to rely on sync calls to do the pushing.

Also:
- make `enqueue_sync` non-async by decoupling table creation and flushing from it
- make table creation lazy (done during actual flushing)
- explicit `flush` method is to be used by the handler

### DIFF
--- a/src/frontend/flight/handler.rs
+++ b/src/frontend/flight/handler.rs
@@ -44,18 +44,18 @@ lazy_static! {
 pub(super) struct SeafowlFlightHandler {
     pub context: Arc<SeafowlContext>,
     pub results: Arc<DashMap<String, Mutex<SendableRecordBatchStream>>>,
-    sync_manager: Arc<RwLock<SeafowlDataSyncWriter>>,
+    sync_writer: Arc<RwLock<SeafowlDataSyncWriter>>,
 }
 
 impl SeafowlFlightHandler {
     pub fn new(
         context: Arc<SeafowlContext>,
-        sync_manager: Arc<RwLock<SeafowlDataSyncWriter>>,
+        sync_writer: Arc<RwLock<SeafowlDataSyncWriter>>,
     ) -> Self {
         Self {
             context: context.clone(),
             results: Arc::new(Default::default()),
-            sync_manager,
+            sync_writer,
         }
     }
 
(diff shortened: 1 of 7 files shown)


src/frontend/flight/mod.rs
```diff
@@ -24,7 +28,14 @@ pub async fn run_flight_server(
...
         .expect("Error parsing the Arrow Flight listen address");
 
-    let handler = SeafowlFlightHandler::new(context);
+    let flush_interval =
+        Duration::from_secs(context.config.misc.sync_conf.flush_task_interval_s);
+    let lock_timeout =
+        Duration::from_secs(context.config.misc.sync_conf.write_lock_timeout_s);
+    let sync_writer = Arc::new(RwLock::new(SeafowlDataSyncWriter::new(context.clone())));
+    let handler = SeafowlFlightHandler::new(context, sync_writer.clone());
+    tokio::spawn(flush_task(flush_interval, lock_timeout, sync_writer));
```
mildbyte, line 37: Do we need to worry about properly terminating this when Seafowl ends / restarting if it crashes / should this use the subsystem setup from the top level that Seafowl uses to run the GC task and other frontends?

src/frontend/flight/handler.rs
```diff
@@ -171,22 +174,24 @@ impl SeafowlFlightHandler {
...
+                sync_writer.enqueue_sync(
+                    log_store,
+                    cmd.sequence_number,
+                    cmd.origin,
+                    sync_schema.expect("Schema available"),
+                    cmd.last,
+                    batches,
+                )?;
+
+                sync_writer.flush().await?;
```
mildbyte, line 192: It's still the case that this `flush()` op won't return until we've had a chance to see if we need to flush and wait for the flush to finish, right? (i.e. Seafowl will never end up using more memory than the limit)

mildbyte (approved):

src/frontend/flight/handler.rs
gruuya: The `flush` call there checks the criteria and while it is met it flushes table after table. The memory usage can only go down inside/after the call.

PR #548, opened 2024-06-26
Make origin a string

eatonphil:
Vadim and Petr have suggested another value I use that will be a string. Since I'm the one that dissuaded us from doing that in the first place, I have offered up a partial attempt at reverting this to what Marko originally implemented.

I have some questions about the tests though. I'm also not sure how to build the protos? I can't see how we do that in CI or anything.

### DIFF
--- a/src/frontend/flight/sync/writer.rs
+++ b/src/frontend/flight/sync/writer.rs
@@ -710,9 +710,7 @@ fn now() -> u64 {
 mod tests {
     use crate::context::test_utils::in_memory_context;
     use crate::frontend::flight::sync::schema::SyncSchema;
-    use crate::frontend::flight::sync::writer::{
-        SeafowlDataSyncWriter, SequenceNumber,
-    };
+    use crate::frontend::flight::sync::writer::{SeafowlDataSyncWriter, SequenceNumber};
     use arrow::{array::RecordBatch, util::data_gen::create_random_batch};
     use arrow_schema::{DataType, Field, Schema, SchemaRef};
     use clade::sync::{ColumnDescriptor, ColumnRole};
@@ -853,7 +851,7 @@ mod tests {
             }
 
             let log_store = ctx.internal_object_store.get_log_store(table_name);
-	    let origin: String = (*origin).to_owned();
+            let origin: String = (*origin).to_owned();
 
             // Determine whether this is the last sync of the sequence, i.e. are there no upcoming
             // syncs with the same sequence number from this origin?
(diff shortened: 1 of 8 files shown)


src/frontend/flight/sync/writer.rs
```diff
@@ -838,12 +838,13 @@ mod tests {
                 sync_mgr.flush_syncs().await.unwrap();
 
                 for (o, durs) in durable_sequences.iter_mut().enumerate() {
-                    let origin = o as i64;
+		    // What is this supposed to mean?
+                    let origin = *o;
```
eatonphil, line 842: I am quite confused about what was going on here. The index in the sequence was being used as the origin id?

src/frontend/flight/sync/writer.rs
```diff
@@ -763,9 +763,9 @@ mod tests {
     const T1: &str = "table_1";
     const T2: &str = "table_2";
     const T3: &str = "table_3";
-    const O1: i64 = 0; // first origin
-    const O2: i64 = 1; // second origin
-    const FLUSH: (&str, i64, i64) = ("__flush", -1, -1);
+    static O1: &str = "1"; // first origin
```
mildbyte, line 766:
```suggestion
    static O1: &str = "0"; // first origin
```

src/frontend/flight/sync/writer.rs
mildbyte: Yeah, it looks like this is just for convenience for this test case (`mut durable_sequences` is parameterized and in the test data the origins have consecutive IDs).

The `to_string()` below makes sense to me.

src/frontend/flight/sync/writer.rs
```diff
@@ -853,19 +854,20 @@ mod tests {
...
             let log_store = ctx.internal_object_store.get_log_store(table_name);
+	    let origin = (*origin).to_string();
 
             // Determine whether this is the last sync of the sequence, i.e. are there no upcoming
             // syncs with the same sequence number from this origin?
             let last = !table_sequence
                 .iter()
                 .skip(sync_no + 1)
-                .any(|&(_, o, s)| *sequence == s && *origin == o);
+                .any(|&(_, o, s)| *sequence == s && o == o);
```
mildbyte, line 864: Now it's comparing `o` to `o` which is always going to be true, I think we're trying to compare the `o` from the outer loop above to the `o` in the lambda

```suggestion
                .any(|&(_, this_o, s)| *sequence == s && o == this_o);
```

eatonphil: Yup, whoops, thank you.

src/frontend/flight/sync/writer.rs
eatonphil: Thanks!

eatonphil:
> I'm also not sure how to build the protos?

I'm still not sure how to build the protos. It seems like I can't run `cargo build` successfully until I do?

eatonphil:
Feel free to squash and merge this too if it is approved.

PR #553, opened 2024-07-02
clippy failing due to not being pinned

eatonphil:
It's nightly so changes to clippy itself means unrelated PRs fail.


eatonphil:
This seems to be passing now, feel free to squash and merge this keeping only the middle commit message tomorrow.

We may want to consider pinning rust versions.

PR #556, opened 2024-07-08
Use the storage location name as the key in the clade schema

gruuya:
As opposed to keying them via their root URL.

### DIFF
--- a/src/catalog/metastore.rs
+++ b/src/catalog/metastore.rs
@@ -97,7 +97,7 @@ impl Metastore {
         let store_options = catalog_schemas
             .stores
             .into_iter()
-            .map(|store| (store.location, store.options))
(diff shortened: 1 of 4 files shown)


src/catalog/metastore.rs
```diff
@@ -140,21 +140,21 @@ impl Metastore {
     async fn build_table(
         &self,
         table: TableObject,
-        store_options: &HashMap<String, HashMap<String, String>>,
+        store_options: &HashMap<String, (String, HashMap<String, String>)>,
```
SergeiPatiakin, line 143: There's 4 Strings in this type and it took me a while to figure out what each of them was. Maybe add a comment about this?

gruuya: Good point, will do.

PR #569, opened 2024-07-16
BEAC-167: Create a crate that instantiates an object store based on URL, settings and environment variables

lizardoluis:
This PR breaks the object store creation logic into a separate crate `object_store_factory`. 

The crate allows building object stores from various configurations and URLs. The key updates include:

- `build_object_store_from_config` function:

  - **Purpose**: Constructs an object store based on a given scheme (Memory, Local, AmazonS3, GoogleCloudStorage) and configuration settings.
  - **Details:** Uses specific modules (memory, local, aws, google) to build the appropriate object store. 

- `build_object_store_from_opts` function:

  - **Purpose:** Constructs an object store based on a URL and associated configuration options.
  - **Details:** Parses the URL to determine the scheme and adjusts options accordingly. Supports Amazon S3 and Google Cloud Storage schemes with specific option modifications.

### DIFF
--- a/object_store_factory/src/aws.rs
+++ b/object_store_factory/src/aws.rs
@@ -100,25 +100,22 @@ pub async fn add_amazon_s3_specific_options(
 pub fn add_amazon_s3_environment_variables(
     options: &mut HashMap<AmazonS3ConfigKey, String>,
 ) {
-    let env_vars = &[
-        ("AWS_ACCESS_KEY_ID", AmazonS3ConfigKey::AccessKeyId),
-        ("AWS_SECRET_ACCESS_KEY", AmazonS3ConfigKey::SecretAccessKey),
-        ("AWS_DEFAULT_REGION", AmazonS3ConfigKey::Region),
-        ("AWS_ENDPOINT", AmazonS3ConfigKey::Endpoint),
-        ("AWS_SESSION_TOKEN", AmazonS3ConfigKey::Token),
-        (
-            "AWS_CONTAINER_CREDENTIALS_RELATIVE_URI",
-            AmazonS3ConfigKey::ContainerCredentialsRelativeUri,
-        ),
-    ];
-
-    for &(env_var_name, config_key) in env_vars.iter() {
-        if let Ok(val) = env::var(env_var_name) {
-            options.insert(config_key, val);
+    for (os_key, os_value) in std::env::vars_os() {
+        if let (Some(key), Some(value)) = (os_key.to_str(), os_value.to_str()) {
+            if key.starts_with("AWS_") {
+                if let Ok(config_key) = key.to_ascii_lowercase().parse() {
+                    options.insert(config_key, value.to_string());
+                }
+            }
         }
     }
 
-    if env::var("AWS_ALLOW_HTTP") == Ok("true".to_string()) {
+    if env::var(
+        AmazonS3ConfigKey::Client(ClientConfigKey::AllowHttp)
+            .as_ref()
+            .to_uppercase(),
+    ) == Ok("true".to_string())
+    {
         options.insert(
             AmazonS3ConfigKey::Client(ClientConfigKey::AllowHttp),
             "true".to_string(),
--- a/object_store_factory/src/aws.rs
+++ b/object_store_factory/src/aws.rs
@@ -1,14 +1,16 @@
+use object_store::ClientConfigKey;
 use object_store::{
     aws::resolve_bucket_region, aws::AmazonS3Builder, aws::AmazonS3ConfigKey,
     ClientOptions, ObjectStore,
 };
 use std::collections::HashMap;
 use std::env;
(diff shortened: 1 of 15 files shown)


object_store_factory/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+use object_store::ObjectStore;
+use object_store::ObjectStoreScheme;
+use std::collections::HashMap;
+use tracing::warn;
+use url::Url;
+
+/// Configuration for object storage, holding settings in a key-value format.
+pub struct Config {
+    pub settings: HashMap<String, String>,
+}
```
gruuya, line 16: Looks like this isn't being used anywhere?

On the other hand if it is, you can declare it as a type alias, since it's a simple wrapper
```rust
pub type StorageConfig = HashMap<String, String>;
```

object_store_factory/src/aws.rs
```diff
@@ -0,0 +1,225 @@
...
+pub fn add_amazon_s3_environment_variables(
+    mut options: HashMap<String, String>,
+) -> HashMap<String, String> {
+    let env_vars = &[
+        ("AWS_ACCESS_KEY_ID", "access_key_id"),
+        ("AWS_SECRET_ACCESS_KEY", "secret_access_key"),
+        ("AWS_DEFAULT_REGION", "region"),
+        ("AWS_ENDPOINT", "endpoint"),
+        ("AWS_SESSION_TOKEN", "session_token"),
+        ("AWS_CONTAINER_CREDENTIALS_RELATIVE_URI", "credentials_uri"),
```
gruuya, line 88: You could use `AmazonS3ConfigKey` for these as well (likewise for other stores).

object_store_factory/src/lib.rs
lizardoluis: I declared it as type alias as sugested. This is used at `build_object_store_from_config`.

src/config/schema.rs
```diff
@@ -119,9 +130,24 @@ pub struct Local {
...
 
+impl InMemory {
+    pub fn to_config(&self) -> HashMap<String, String> {
+        // InMemory does not have any specific configurations.
+        HashMap::new()
+    }
+}
+
 #[derive(Deserialize, Debug, PartialEq, Eq, Clone)]
 pub struct S3 {
```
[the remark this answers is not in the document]
lizardoluis: @gruuya This change will be a snow ball as these structs are used in several part of the code. Would be ok to address this change in a separate PR, so that we reduce the complexity of it?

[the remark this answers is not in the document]
gruuya: Yup, sounds good.

object_store_factory/src/aws.rs
```diff
@@ -0,0 +1,322 @@
...
+    let env_vars = &[
+        ("AWS_ACCESS_KEY_ID", AmazonS3ConfigKey::AccessKeyId),
+        ("AWS_SECRET_ACCESS_KEY", AmazonS3ConfigKey::SecretAccessKey),
+        ("AWS_DEFAULT_REGION", AmazonS3ConfigKey::Region),
+        ("AWS_ENDPOINT", AmazonS3ConfigKey::Endpoint),
+        ("AWS_SESSION_TOKEN", AmazonS3ConfigKey::Token),
+        (
+            "AWS_CONTAINER_CREDENTIALS_RELATIVE_URI",
+            AmazonS3ConfigKey::ContainerCredentialsRelativeUri,
+        ),
```
gruuya, line 112: Seems like the first elements are redundant, since you can get them from the second element(s): `config_key.as_ref().to_uppercase()`

object_store_factory/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+///
+/// # Returns
+///
+/// Returns a `Result` that, on success, contains a boxed `ObjectStore` trait object. On failure, it returns an `object_store::Error`
+/// indicating what went wrong, such as an unsupported URL scheme.
+///
+/// # Errors
+///
+/// * If the URL scheme is unsupported or there is an error parsing the URL or options, an error is returned.
+pub async fn build_object_store_from_opts(
```
mildbyte, line 75: This should also support building a local FS object store using the `file://` protocol

gruuya: Should be resolve now with 26bb61effb9d8bf70f7a90f4b2f390b45b7547ae.

gruuya (approved):

object_store_factory/src/aws.rs
lizardoluis: I changed it completely to be more robust and accept more variables. I implemented something similar to their `from_env()` algorithm like here https://docs.rs/object_store/latest/src/object_store/aws/builder.rs.html#429-443.

gruuya: Nice!

PR #580, opened 2024-07-30
BEAC-167: Move object store configs to object_store_factory

lizardoluis:
This PR moves the object store configuration structs into the new crate `object_store_factory`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,8 @@ tracing-log = "0.2"
 tracing-subscriber = { version = "0.3.18", features = ["json", "env-filter"] }
 url = "2.5"
 
+
+
 [patch.crates-io]
 # Pick up backport of https://github.com/apache/arrow-datafusion/pull/11386
 datafusion = { git = "https://github.com/splitgraph/arrow-datafusion", branch = "backport-pr11386" }
@@ -65,11 +67,9 @@ include = [
 
 [features]
 catalog-postgres = ["sqlx/postgres"]
-default = ["catalog-postgres", "frontend-arrow-flight", "frontend-postgres", "object-store-s3", "object-store-gcs", "remote-tables"]
+default = ["catalog-postgres", "frontend-arrow-flight", "frontend-postgres", "remote-tables"]
 frontend-arrow-flight = ["dep:tonic", "dep:arrow-flight", "arrow-flight/flight-sql-experimental"]
 frontend-postgres = ["convergence", "convergence-arrow"]
-object-store-gcs = ["object_store/gcp"]
-object-store-s3 = ["object_store/aws"]
 remote-tables = ["dep:datafusion-remote-tables"]
 
 [dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -128,7 +128,7 @@ rmp = "0.8.11"
 rmp-serde = "1.1.1"
 rmpv = { version = "1.0.0", features = ["with-serde"] }
 rustyline = "13.0"
-serde = "1.0.156"
+serde = { workspace = true }
 serde_json = "1.0.93"
 sha2 = ">=0.10.1"
 sqlparser = { version = "0.47", features = ["visitor"] }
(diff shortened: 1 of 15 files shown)


Cargo.toml
```diff
@@ -22,6 +22,8 @@ futures = "0.3"
 
 itertools = ">=0.10.0"
 object_store = { version = "0.10.2", features = ["aws", "azure", "gcp"] }
+
+serde = "1.0.156"
```
gruuya, line 26: We should use this in `[dependencies]` below.

src/config/schema.rs
```diff
@@ -52,30 +53,6 @@ pub struct SeafowlConfig {
...
 
-#[derive(Deserialize, Debug, PartialEq, Eq, Clone)]
-#[serde(tag = "type", rename_all = "snake_case")]
-pub enum ObjectStore {
-    Local(Local),
-    #[serde(rename = "memory")]
-    InMemory(InMemory),
-    #[cfg(feature = "object-store-s3")]
-    S3(S3),
-    #[cfg(feature = "object-store-gcs")]
```
gruuya, line 63: We should probably get rid of these two features in Cargo.toml (not used in the new configs, and I don't think we ever opt-out of these).

src/config/schema.rs
```diff
@@ -923,6 +786,7 @@ cache_control = "private, max-age=86400"
     fn test_parse_config_invalid_s3() {
         let error =
             load_config_from_string(TEST_CONFIG_INVALID_S3, false, None).unwrap_err();
+        println!("##### ERROR: {:?}", error.to_string());
```
gruuya, line 789:
```suggestion
```

src/object_store/wrapped.rs
```diff
@@ -89,16 +85,15 @@ impl InternalObjectStore {
...
+        let prefix = match self.config.clone() {
+            ObjectStoreConfig::AmazonS3(aws_config) => aws_config.prefix,
+            ObjectStoreConfig::GoogleCloudStorage(google_config) => google_config.prefix,
+            _ => None,
+        };
+
+        match prefix {
+            Some(prefix) => Path::from(format!("{prefix}/{table_prefix}")),
+            None => Path::from(table_prefix),
         }
```
gruuya, line 97: Seems simpler to do it with one pattern match as before?

```suggestion
        match self.config.clone() {
            ObjectStoreConfig::AmazonS3(S3Config {
                prefix: Some(prefix),
                ..
            })
            | ObjectStoreConfig::GoogleCloudStorage(GCSConfig {
                prefix: Some(prefix),
                ..
            }) => Path::from(format!("{prefix}/{table_prefix}")),
            _ => Path::from(table_prefix),
        }
```

gruuya (approved):

Cargo.toml
lizardoluis: @gruuya I get an error if I do not set it here, as I'm inheriting it in the `object_store_factory` Cargo.toml. Btw, it is also set in the `[dependecies]`.

The error is: 

```
error: failed to load manifest for workspace member `/Users/lizardo/Workspace/seafowl/object_store_factory`
referenced by workspace at `/Users/lizardo/Workspace/seafowl/Cargo.toml`

Caused by:
  failed to parse manifest at `/Users/lizardo/Workspace/seafowl/object_store_factory/Cargo.toml`

Caused by:
  error inheriting `serde` from workspace root manifest's `workspace.dependencies.serde`

Caused by:
  `dependency.serde` was not found in `workspace.dependencies
```

src/config/schema.rs
lizardoluis: Missed that

src/object_store/wrapped.rs
lizardoluis: I reverted, though I think the readability was better with the change.

Cargo.toml
gruuya: No, i meant for serde to remain in `workspace.dependencies`, just to change line 129/131 in this file to inherit from it `serde = { workspace = true }`.

src/config/schema.rs
lizardoluis: I purged also some occurrences of it in the code.

Cargo.toml
lizardoluis: Ahh, got it.

PR #581, opened 2024-07-30
Optional sync sequence number

gruuya:
Make sequence numbers optional in the protocol, and don't rely on their presence for the flushing mechanism to work. The reason for this is that PGD might not know the sequence number until the very end (last message) of the transaction.

This is achieved with a single queue of pending transactions, whereby with bookkeeping we:

1. track which txs are complete (as soon as they have `Some` sequence number)
2. track which txs have all tables/locations flushed out
3. make sure to mark a tx as durable only if all preceding txs are also durable

### DIFF
--- a/clade/proto/sync.proto
+++ b/clade/proto/sync.proto
@@ -35,13 +35,9 @@ message DataSyncCommand {
   // Integer identifying data source
   string origin = 4;
 
-  // Monotonically-increasing transaction number
-  uint64 sequence_number = 5;
-
-  // True if this is the last command in the transaction denoted
-  // by this sequence number which might then become durable after
-  // the next flush.
-  bool last = 6;
+  // Monotonically-increasing transaction number.
+  // Only specified in the last message of a transaction
+  optional uint64 sequence_number = 5;
 }
 
 message DataSyncResult {
--- a/src/frontend/flight/sync/writer.rs
+++ b/src/frontend/flight/sync/writer.rs
@@ -122,7 +122,7 @@ pub(super) struct DataSyncCollection {
 }
 
 // An object corresponding to a single `do_put` call.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub(super) struct DataSyncItem {
     // The (internal) id of the transaction that this change belongs to; for now it corresponds to
     // a random v4 UUID generated for the first message received in this transaction.
@@ -571,21 +571,15 @@ impl SeafowlDataSyncWriter {
                     ..
                 }) = self.txs.get(&sync.tx_id)
                 {
-                    Some(SyncCommitInfo {
-                        version: 1,
-                        origin: origin.clone(),
-                        sequence: *seq,
-                        new_tx,
-                    })
+                    Some(SyncCommitInfo::new(origin, *seq).with_new_tx(new_tx))
                 } else {
                     None
                 }
             })
-            .or(last_sync_commit.clone().map(|mut sync_commit| {
+            .or(last_sync_commit.clone().map(|sync_commit| {
                 // Inherit the previous full commit identifiers, but update the fact that we're now
                 // starting a new transaction too.
-                sync_commit.new_tx = new_tx;
-                sync_commit
+                sync_commit.with_new_tx(new_tx)
             }));
 
         (syncs, new_sync_commit)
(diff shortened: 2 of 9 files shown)


src/frontend/flight/sync/writer.rs
```diff
@@ -770,48 +844,62 @@ mod tests {
...
+            vec![None, Some(200)],
+        ]
+    )]
+    #[should_panic(
+        expected = "Transaction from a new origin (origin-B) started without finishing the transaction from the previous one (origin-A)"
+    )]
+    #[case::incomplete_transaction(
+        &[(T1, A, None), (T2, B, Some(100)), FLUSH],
+        vec![],
+    )]
```
gruuya, line 867: @eatonphil I'm working under the assumption that this isn't possible (hence the error).

eatonphil: Perhaps it should be ignored rather than panic? Panicing is fine for tests but while this isn't a "valid" scenario it is one that may actually happen.

gruuya: Oh it's not actually a panic, or to be more specific it is in a test, but in practice it just responds with an Error to the rpc call.

eatonphil:
Ok so we **must not** fill out the sequence number until the final batch for the transaction.

eatonphil:
(Sorry I meant for that to be a question. :D )

gruuya:
> Ok so we **must not** fill out the sequence number until the final batch for the transaction.

Well in principle I could bring back the `last` field, and have that denote the actual last message in the transaction. This would somewhat loosen the requirements on the sequence number, i.e. it could also become non-None prior to the last message. So something like the following sequence of messages:
```rust
(table_1, origin-A, None, false) <- still don't know the sequence number for this transaction
(table_2, origin-A, Some(1000), false) <- here you find out the sequence number and can start passing it along
(table_3, origin-A, Some(1000), true) <- the transaction is now complete (last = true)
```

Then we'd require that at least one of the messages in the transaction contains the sequence number.

Let me know if this would simplify things for you somehow.

eatonphil:
No I think it seems simpler to do it as you have it. I just wanted to confirm

clade/proto/sync.proto
```diff
@@ -35,13 +35,9 @@ message DataSyncCommand {
...
-  // Monotonically-increasing transaction number
-  uint64 sequence_number = 5;
-
-  // True if this is the last command in the transaction denoted
-  // by this sequence number which might then become durable after
-  // the next flush.
-  bool last = 6;
+  // Monotonically-increasing transaction number.
+  // Only specified in the last message of a transaction
+  optional uint64 sequence_number = 5;
```
mildbyte, line 40: (just noting this won't be bw compatible with existing clients)

src/frontend/flight/sync/writer.rs
```diff
@@ -158,36 +161,54 @@ impl SeafowlDataSyncWriter {
...
+        let tx_id = if let Some((tx_id, tx)) = self.txs.last_mut()
+            && tx.sequence.is_none()
+        {
+            if origin != tx.origin {
+                return Err(SyncError::InvalidMessage {
+                    reason: format!(
+                        "Transaction from a new origin ({origin}) started without finishing the transaction from the previous one ({})",
+                        tx.origin
+                    )
+                });
```
mildbyte, line 181: This can happen during failover though? @eatonphil

src/frontend/flight/sync/writer.rs
```diff
@@ -465,32 +491,98 @@ impl SeafowlDataSyncWriter {
...
                 _ => None,
             },
         )
     }
 
+    // Given a known `SyncCommitInfo` from a previous commit try to deduce whether some of the pending
+    // sync are actually redundant (since they were already persisted) and skip them.
+    // Also create a new `SyncCommitInfo` that corresponds to the last full transaction in the
+    // pending changes,
+    fn skip_syncs<'a>(
```
mildbyte, line 521: Didn't @eatonphil mention something about how we can't actually skip over messages with an LSN < last committed because they might have changed?

mildbyte (approved):

src/frontend/flight/sync/writer.rs
eatonphil: Yes I think I commented about this in tests Marko mentioned me on. You should not panic. While it's an invalid scenario it is certainly a possible scenario. You should not crash or become unrecoverable from this.

eatonphil: Another example is not even failover but just if the writer process itself crashes midway through. We may not send the full transaction and then may send it again (from the same origin).

gruuya: Note that the above merely returns an error to the caller, and doesn't panic/crash the process. Then presumably the other side can re-adjust it's position in the message queue and try again?

PR #591, opened 2024-08-06
Support empty syncs with no batches provided as payload

gruuya:
This is to signal a transaction end.

Also rename batch compaction to squashing throughout.

### DIFF
--- a/src/frontend/flight/sql.rs
+++ b/src/frontend/flight/sql.rs
@@ -3,6 +3,7 @@ use crate::frontend::flight::handler::{
     SeafowlFlightHandler, SEAFOWL_SQL_DATA, SEAFOWL_SYNC_CALL_MAX_ROWS,
 };
 use crate::frontend::flight::sync::schema::SyncSchema;
+use crate::frontend::flight::sync::SyncError;
 use arrow::record_batch::RecordBatch;
 use arrow_flight::decode::FlightRecordBatchStream;
 use arrow_flight::encode::FlightDataEncoderBuilder;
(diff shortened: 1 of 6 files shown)


src/frontend/flight/sql.rs
```diff
@@ -211,27 +212,30 @@ impl FlightSqlService for SeafowlFlightHandler {
...
+                cmd.column_descriptors.clone(),
+                batches.first().unwrap().schema(),
             )
+            .map_err(|err| {
+                warn!("{err}");
+                Status::invalid_argument(err.to_string())
+            })?
         } else {
-            None
+            SyncSchema::empty()
```
mildbyte, line 224: Can this actually legitimately happen now or should we always expect that there is a schema?

mildbyte (approved):

src/frontend/flight/sql.rs
gruuya: It seems that providing an empty payload (`vec![]`) is easier than providing a non-empty payload with empty record batches (`vec![RecordBatch::new_empty(schema)]`), so I think yes.

Let's hear from @eatonphil though.

eatonphil: > It seems that providing an empty payload (vec![]) is easier than providing a non-empty payload with empty record batches (vec![RecordBatch::new_empty(schema)]), so I think yes.

Well, let me try. I have such a hard time with the Arrow APIs I was scared to try. But if you think that's better I'll give it a shot.

gruuya: It may still better of to have special casing of the empty payload, than to rely on the rest of the mechanism to adequately process those messages (so this change is for the better anyway).

eatonphil: Alright, let's go with it then.

PR #592, opened 2024-08-06
Refactor `object_store_factory` and create `StorageLocationInfo`

lizardoluis:
This PR refactors the `object_store_factory` with the changes:

- Added `allow_http` and `skip_signature` to `S3Config`;
- Changed the `build_object_store(ObjectStoreConfig)` to be a implementation of `ObjectStoreConfig `. 
- Implemented the struct `StorageLocationInfo` with the method `build_storage_location_info`.


object_store_factory/src/aws.rs
```diff
@@ -49,38 +77,116 @@ impl S3Config {
...
     }
 
-    if let (Some(access_key_id), Some(secret_access_key)) =
-        (&config.access_key_id, &config.secret_access_key)
-    {
-        builder = builder
-            .with_access_key_id(access_key_id.clone())
-            .with_secret_access_key(secret_access_key.clone());
+    pub fn build_amazon_s3(&self) -> Result<Arc<dyn ObjectStore>, object_store::Error> {
+        assert!(self.allow_http, "allow_http must be true for Amazon S3");
```
gruuya, line 148: This should be the case only for testing on e.g. MinIO, otherwise it should be disallowed (no point in `allow_http` field if this assert is here).

gruuya (approved):

object_store_factory/src/aws.rs
lizardoluis: I forgot to remove this. I wasn't sure if the other code at any point calls this with `allow_http == false`, so I added this to catch. I will remove

gruuya: Ah makes sense, thanks.

PR #593, opened 2024-08-07
Add flag denoting first sync response

gruuya:
The flag is tracked in `SeafowlFlightHandler` since it is the one that actually construct the response (not `SeafowlDataSyncWriter`), and in addition it has short-circuiting for the "probing" request.

### DIFF
--- a/src/frontend/flight/handler.rs
+++ b/src/frontend/flight/handler.rs
@@ -4,13 +4,14 @@ use arrow::record_batch::RecordBatch;
 use arrow_flight::sql::metadata::{SqlInfoData, SqlInfoDataBuilder};
 use arrow_flight::sql::{ProstMessageExt, SqlInfo, TicketStatementQuery};
 use arrow_flight::{FlightDescriptor, FlightEndpoint, FlightInfo, Ticket};
-use clade::sync::{DataSyncCommand, DataSyncResult};
+use clade::sync::{DataSyncCommand, DataSyncResponse};
 use dashmap::DashMap;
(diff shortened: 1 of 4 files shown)


src/frontend/flight/handler.rs
```diff
@@ -47,6 +48,8 @@ pub(super) struct SeafowlFlightHandler {
     pub context: Arc<SeafowlContext>,
     pub results: Arc<DashMap<String, Mutex<SendableRecordBatchStream>>>,
     sync_writer: Arc<RwLock<SeafowlDataSyncWriter>>,
+    // Denotes whether we're past the first sync reponse, thus indicating Seafowl (re)starts
```
mildbyte, line 51:
```suggestion
    // Denotes whether we're past the first sync response, thus indicating Seafowl (re)starts
```

src/frontend/flight/handler.rs
```diff
@@ -47,6 +48,8 @@ pub(super) struct SeafowlFlightHandler {
...
+    first_sync: Arc<AtomicBool>,
```
mildbyte, line 52: This doesn't need to be tracked for each origin separately?

mildbyte (approved):

src/frontend/flight/handler.rs
gruuya: My understanding was no, the overall response order is important, cc @eatonphil?

eatonphil: That's right

PR #607, opened 2024-08-13
Allow running Seafowl without an object store / catalog config.

mildbyte:
In this case, Seafowl expects the metastore and the object store settings to be passed to it over the Flight RPC endpoint (when querying or writing) and will error out otherwise.

This required some weird refactoring because so many things assume there is a "default" object store and a metastore, so all SeafowlContext methods that want to access the internal default object store for writes now do it through a method that errors out if it's not set up. For the metastore, this change defines an EmptyMetastore that errors if Seafowl actually tries to use it.

### DIFF
--- a/src/context/physical.rs
+++ b/src/context/physical.rs
@@ -8,6 +8,7 @@ use crate::nodes::{
 };
 use crate::object_store::factory::build_object_store;
 use crate::object_store::http::try_prepare_http_url;
+use crate::object_store::wrapped::InternalObjectStore;
 use crate::provider::project_expressions;
 use crate::utils::gc_databases;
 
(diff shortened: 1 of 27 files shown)


src/context/physical.rs
```diff
@@ -94,106 +104,11 @@ impl SeafowlContext {
...
             // mechanism in our case too.
             LogicalPlan::Ddl(DdlStatement::CreateExternalTable(
-                cmd @ CreateExternalTable {
-                    ref name,
-                    ref location,
-                    ..
-                },
+                cmd @ CreateExternalTable { .. },
             )) => {
-                // Replace the table name with the fully qualified one that has our staging schema
```
mildbyte, line 103: This was refactored out into `prepare_create_external_table` without any logic changes

PJMODOS:
So how do I use this, do I simply skip [catalog] part of config?

gruuya (approved):

mildbyte:
@PJMODOS Yeah, both catalog and object_store can now be skipped completely

PR #616, opened 2024-08-16
Granular sync pruning

gruuya:
Make sync pruning as granular as possible by testing each PK row in all pending sync messages for possible membership in each partition, and do so in a vectorized way (for sync rows).

Possible PK row membership is done in a conjunctive manner—each PK column of the row is tested for range in each partition and the overall result is obtained by AND-ing together individual tests across the columns.


mildbyte (approved):

gruuya:
Closing in favor of #622.

PR #629, opened 2024-08-24
add copy and delete logic for os error 38 

mrchypark:
I got error when create table on localfilesystem.

```
Object Store error: Generic LocalFileSystem error: Unable to copy file from /data/3a6b2ef0-fcf6-4bc2-a75a-ad2714c3ae69/_delta_log/_commit_bdeba02b-7620-42d7-8ea8-34db0b7a0f2c.json.tmp to /data/3a6b2ef0-fcf6-4bc2-a75a-ad2714c3ae69/_delta_log/00000000000000000000.json: Function not implemented (os error 38)
```

I use aks and azure blob csi driver.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -62,6 +62,11 @@ jobs:
           export PATH=$PATH:$HOME/d/protoc/bin
           cargo build --release
 
+      - name: Test invoking the binaries
+        shell: bash
+        run: |
+          ./target/release/seafowl --version || exit 1
+
       - name: Docker meta
         id: docker_meta
         uses: docker/metadata-action@v5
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly.yml
```diff
@@ -14,17 +14,17 @@ jobs:
   build_binary:
```
mildbyte, line 14: Thanks! Changes to `utils.rs` look good, but can you revert changes to this CI file before we merge?

mrchypark:
Hi! I've found that this PR doesn't solve the problem and identified the root cause. I'll organize it into a separate issue and PR. Thank you for your attention to this matter.

PR #630, opened 2024-08-26
Use the new public method for getting Add stats from the logs

gruuya:
Instead of the previous one which is empty for adds obtained via `DeltaTableState::file_actions`, use the from https://github.com/delta-io/delta-rs/pull/2822.

This allows us to trigger the fine-grained pruning routine.

### DIFF
--- a/object_store_factory/src/aws.rs
+++ b/object_store_factory/src/aws.rs
@@ -1,8 +1,8 @@
-use object_store::aws::S3ConditionalPut;
-use object_store::{
-    aws::resolve_bucket_region, aws::AmazonS3Builder, aws::AmazonS3ConfigKey, path::Path,
-    ClientConfigKey, ClientOptions, ObjectStore,
+use object_store::aws::{
+    resolve_bucket_region, AmazonS3Builder, AmazonS3ConfigKey, S3ConditionalPut,
 };
+use object_store::path::Path;
+use object_store::{ClientConfigKey, ClientOptions, ObjectStore};
 use serde::Deserialize;
 use std::collections::HashMap;
 use std::env;
--- a/object_store_factory/src/aws.rs
+++ b/object_store_factory/src/aws.rs
@@ -211,6 +211,10 @@ pub async fn add_amazon_s3_specific_options(
         let region = detect_region(url).await.unwrap();
         options.insert(AmazonS3ConfigKey::Region, region.to_string());
     }
+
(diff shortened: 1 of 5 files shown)


object_store_factory/src/aws.rs
```diff
@@ -141,7 +142,8 @@ impl S3Config {
         let mut builder = AmazonS3Builder::new()
             .with_region(self.region.clone().unwrap_or_default())
             .with_bucket_name(self.bucket.clone())
-            .with_allow_http(self.allow_http);
+            .with_allow_http(self.allow_http)
+            .with_conditional_put(S3ConditionalPut::ETagMatch);
```
gruuya, line 146: This was needed since otherwise we hit this error in object_store: https://github.com/apache/arrow-rs/blob/855666d9e9283c1ef11648762fe92c7c188b68f1/object_store/src/aws/mod.rs#L171

We hit these since with the delta-rs bump we picked up https://github.com/delta-io/delta-rs/pull/2813, and there the put mode is explicitly being set to `PutMode::Create` (whereas previously it defaulted to `PutMode::Overwrite`).

lizardoluis (approved):

object_store_factory/src/aws.rs
```diff
@@ -1,3 +1,4 @@
+use object_store::aws::S3ConditionalPut;
 use object_store::{
     aws::resolve_bucket_region, aws::AmazonS3Builder, aws::AmazonS3ConfigKey, path::Path,
     ClientConfigKey, ClientOptions, ObjectStore,
```
lizardoluis, line 4: Just to keep this more organized:

```suggestion
use object_store::aws::{AmazonS3Builder, AmazonS3ConfigKey, S3ConditionalPut, resolve_bucket_region};
use object_store::{ ClientConfigKey, ClientOptions, ObjectStore };
use object_store::path::Path;
```

object_store_factory/src/aws.rs
lizardoluis: If we always need to add this when creating a S3 object storage, then I think you will also need to add it to the `add_amazon_s3_specific_options()` function. Because object storage can also be created using this factory with options.

object_store_factory/src/aws.rs
gruuya: Thanks, good catch.

PR #639, opened 2024-08-27
Add support for TRUNCATE command

SergeiPatiakin:
## What
- Add support for `TRUNCATE TABLE x`
- This is equivalent to `DELETE FROM x`, but it also works with inline metastore

## How
- Skip updating `table_versions` to ensure it works with inline metastore

### DIFF
--- a/src/context/physical.rs
+++ b/src/context/physical.rs
@@ -654,26 +654,27 @@ impl SeafowlContext {
                         SeafowlExtensionNode::Truncate(Truncate {
                             table_name, ..
                         }) => {
-                            let mut table =
-                                self.try_get_delta_table(table_name.clone()).await?;
+                            let mut table = self.try_get_delta_table(table_name).await?;
                             table.load().await?;
                             let snapshot = table.snapshot()?;
-                            let removes = snapshot.file_actions()?;
                             let deletion_timestamp = SystemTime::now()
                                 .duration_since(UNIX_EPOCH)
                                 .unwrap()
                                 .as_millis()
                                 as i64;
-                            let actions: Vec<Action> = removes
+                            let actions: Vec<Action> = snapshot
+                                .file_actions()?
                                 .into_iter()
-                                .map(|remove| {
+                                .map(|add_action| {
                                     Action::Remove(Remove {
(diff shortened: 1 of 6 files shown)


gruuya (approved):

src/context/physical.rs
```diff
@@ -651,6 +651,42 @@ impl SeafowlContext {
                                 .await?;
                             Ok(make_dummy_exec())
                         }
+                        SeafowlExtensionNode::Truncate(Truncate {
+                            table_name, ..
+                        }) => {
+                            let mut table =
+                                self.try_get_delta_table(table_name.clone()).await?;
```
lizardoluis, line 658: Please remove the unnecessary clone of `table_name`, if it works.

```suggestion
                            let mut table =
                                self.try_get_delta_table(table_name).await?;
```

src/context/physical.rs
```diff
@@ -651,6 +651,42 @@ impl SeafowlContext {
...
+                            table.load().await?;
+                            let snapshot = table.snapshot()?;
+                            let removes = snapshot.file_actions()?;
```
lizardoluis, line 661: For clarity:

```suggestion
                            let file_actions = snapshot.file_actions()?;
```

src/datafusion/parser.rs
```diff
@@ -219,6 +223,20 @@ impl<'a> DFParser<'a> {
         })))
     }
 
+    pub fn parse_truncate(&mut self) -> Result<Statement, ParserError> {
+        let table_name: ObjectName = if self.parser.parse_keyword(Keyword::TABLE) {
+            self.parser.parse_object_name(true)?
+        } else {
+            return self.expected("TABLE as a TRUNCATE target", self.parser.peek_token());
+        };
```
lizardoluis, line 231: You can handle the error first here to simplify the flow and reduce the need for nested code.

```suggestion
    if !self.parser.parse_keyword(Keyword::TABLE) {
        return self.expected("TABLE as a TRUNCATE target", self.parser.peek_token());
    }

    let table_name = self.parser.parse_object_name(true)?;
```

lizardoluis (approved):

PR #640, opened 2024-08-28
Logical inter-sync squashing

gruuya:
# What
Implement logical inter-sync squashing during flushing in the `SeafowlDataSyncWriter`, such that the resulting stream is the equivalent combined/total sync batch to be applied to the base scan.

# How
Instead of starting of with a base scan of parquet files to re-write, and applying each sync batch iteratively on top of it, flip it around and first do logical squashing of sync batches and only then do a single join with the base scan.

# Why
- This greatly improves the efficiency of the plan execution and therefore writes at the expense of code complexity. 
- The perf improvement is especially pronounced in case of UPDATEs that touch many partition files, since we avoid doing K joins for N base rows when applying each sync. 
- Instead we have only 1 join with N base rows, whereas the sync joins grow in size linearly (up to `max_syncs_per_url` * `SEAFOWL_SYNC_CALL_MAX_ROWS` ~= 3 M rows by deafult).
- This will also be useful when we transition from CoW to MoR, since we'll need sync-squashing to be able to dump the super-sync batch to Delta Lake (and which the optimizer/compactor can then unravel into proper Adds)

# Tests
Present ones are passing, but needs new ones too

### DIFF
--- a/src/frontend/flight/sync/utils.rs
+++ b/src/frontend/flight/sync/utils.rs
@@ -426,109 +426,24 @@ pub(super) fn merge_schemas(
     // TODO: implement sync role indexing in SyncSchema to make all the lookups below cheap
     // Project all sync columns from the lower schema
     for lower_sync_col in lower_schema.columns() {
-        let maybe_upper_sync_col =
-            upper_schema.column(lower_sync_col.name(), lower_sync_col.role());
-
-        // Build the expression that is going to be used when lower sync column is missing
-        let otherwise = match lower_sync_col.role() {
-            ColumnRole::OldPk => {
-                let upper_old_pk_col = maybe_upper_sync_col
-                    .expect("Old/New Pk columns must be the same across syncs");
-
-                when(
-                    col(LOWER_SYNC).is_null(),
-                    col(Column::new(
-                        Some(UPPER_SYNC),
-                        upper_old_pk_col.field().name(),
-                    )),
-                )
-                .otherwise(col(Column::new(
-                    Some(LOWER_SYNC),
-                    lower_sync_col.field().name(),
-                )))?
-            }
-            ColumnRole::NewPk => {
-                let upper_new_pk_col = maybe_upper_sync_col
-                    .expect("Old/New Pk columns must be the same across syncs");
-
-                col(Column::new(
-                    Some(UPPER_SYNC),
-                    upper_new_pk_col.field().name(),
-                ))
-            }
+        // Build the expression to project the combined column
+        let expr = match lower_sync_col.role() {
+            ColumnRole::OldPk => project_old_pk_column(lower_sync_col, upper_schema)?,
+            ColumnRole::NewPk => project_new_pk_column(lower_sync_col, upper_schema)?,
             ColumnRole::Changed => {
-                if let Some(upper_changed_col) = maybe_upper_sync_col {
-                    // The upper schema has the corresponding `Changed` column, take its values
-                    col(Column::new(
-                        Some(UPPER_SYNC),
-                        upper_changed_col.field().name(),
-                    ))
-                } else if upper_schema
-                    .column(lower_sync_col.name(), ColumnRole::Value)
-                    .is_some()
-                {
-                    // The upper schema doesn't have the `Changed` column at hand, but it does have
-                    // the corresponding `Value` column, project true to pick those values up
-                    lit(true)
-                } else {
-                    // The upper sync has neither the `Changed` nor the associated `Value` column,
-                    // project false to avoid overriding NULL values
-                    lit(false)
-                }
+                project_lower_changed_column(lower_sync_col, upper_schema)?
             }
-            ColumnRole::Value => {
-                if maybe_upper_sync_col.is_none()
-                    && lower_schema
-                        .column(lower_sync_col.name(), ColumnRole::Changed)
-                        .is_none()
-                {
-                    // There's no corresponding `Value` sync column in the upper schema, and
-                    // there's no `Changed` column in the lower schema either, we need to project
-                    // the `Changed` column that takes all the values from the lower schema and
-                    // no value from the upper one
-                    col_desc.insert(ColumnDescriptor {
-                        role: ColumnRole::Changed as _,
-                        name: lower_sync_col.name().clone(),
-                    });
-                    projection.push(
-                        when(col(LOWER_SYNC).is_null(), lit(false))
-                            .otherwise(lit(true))?,
-                    );
-                }
-
-                if let Some(upper_value_col) = maybe_upper_sync_col {
-                    col(Column::new(
-                        Some(UPPER_SYNC),
-                        upper_value_col.field().name(),
-                    ))
-                } else {
-                    // The upper sync doesn't have this column but the lower sync already has
-                    // the `Changed` column which was/will be included in the output.
-                    // Project NULLs in its place.
-                    when(
-                        col(UPPER_SYNC).is_null(),
-                        lit(ScalarValue::Null
-                            .cast_to(lower_sync_col.field().data_type())?),
-                    )
-                    .otherwise(col(Column::new(
-                        Some(LOWER_SYNC),
-                        lower_sync_col.field().name(),
-                    )))?
-                }
-            }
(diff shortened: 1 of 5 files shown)


src/frontend/flight/sync/utils.rs
```diff
@@ -414,6 +415,206 @@ pub(super) fn get_prune_map(
...
+            name: lower_sync_col.name().clone(),
+        });
+        projection.push(
+            when(
+                col(UPPER_SYNC).is_null(),
+                col(Column::new(Some(LOWER_SYNC), lower_sync_col.field().name())),
+            )
+            .otherwise(otherwise)?,
+        );
+    }
```
lizardoluis, line 532: This whole loop can be modularized like this. Which can be reused in the loop below for the `upper_sync_col`.  (Please verify if I'm not breaking the code). 

```suggestion
    for lower_sync_col in lower_schema.columns() {
        let expr = match lower_sync_col.role() {
            ColumnRole::OldPk | ColumnRole::NewPk => project_pk_columns(lower_sync_col, upper_schema),
            ColumnRole::Changed => project_changed_columns(lower_sync_col, upper_schema),
            ColumnRole::Value => project_value_columns(lower_sync_col, lower_schema, upper_schema),
        }?;
        projection.push(expr);

        col_desc.insert(ColumnDescriptor {
            role: lower_sync_col.role() as _,
            name: lower_sync_col.name().clone(),
        });
    }
```

With: 

```
fn project_pk_columns(
    sync_col: &SyncColumn,
    upper_schema: &SyncSchema,
) -> SyncResult<Expr> {
    let upper_sync_col = upper_schema
        .column(sync_col.name(), sync_col.role())
        .expect("Old/New Pk columns must be the same across syncs");

    Ok(when(
        col(LOWER_SYNC).is_null(),
        col(Column::new(Some(UPPER_SYNC), upper_sync_col.field().name())),
    )
    .otherwise(col(Column::new(Some(LOWER_SYNC), sync_col.field().name())))?)
}

fn project_changed_columns(
    sync_col: &SyncColumn,
    other_schema: &SyncSchema,
) -> SyncResult<Expr> {
    let upper_changed_col = other_schema.column(sync_col.name(), sync_col.role());

    Ok(if upper_changed_col.is_some() {
        col(Column::new(Some(UPPER_SYNC), upper_changed_col.field().name()))
    } else if other_schema.column(sync_col.name(), ColumnRole::Value).is_some() {
        lit(true)
    } else {
        lit(false)
    })
}

fn project_value_columns(
    sync_col: &SyncColumn,
    lower_schema: &SyncSchema,
    upper_schema: &SyncSchema,
) -> SyncResult<Expr> {
    let upper_value_col = upper_schema.column(sync_col.name(), ColumnRole::Value);

    Ok(if upper_value_col.is_some() {
        col(Column::new(Some(UPPER_SYNC), upper_value_col.field().name()))
    } else {
        when(
            col(UPPER_SYNC).is_null(),
            lit(ScalarValue::Null.cast_to(sync_col.field().data_type())?),
        )
        .otherwise(col(Column::new(Some(LOWER_SYNC), sync_col.field().name())))?
    })
}
```

src/frontend/flight/sync/utils.rs
```diff
@@ -414,6 +415,206 @@ pub(super) fn get_prune_map(
...
+    // subsequent sync merge
+    let projection = col_desc
+        .iter()
+        .zip(projection)
+        .map(|(cd, expr)| {
+            expr.alias_qualified(Some(LOWER_SYNC), SyncColumn::canonical_field_name(cd))
+        })
+        .collect();
+
+    Ok((col_desc, projection))
```
lizardoluis, line 615:
```suggestion
    for upper_sync_col in upper_schema.columns() {
        if col_desc.contains(&ColumnDescriptor {
            role: upper_sync_col.role() as _,
            name: upper_sync_col.name().clone(),
        }) {
            continue; // Column already projected
        }

        let expr = match upper_sync_col.role() {
            ColumnRole::Changed => project_changed_columns(upper_sync_col, lower_schema),
            ColumnRole::Value => project_value_columns(upper_sync_col, lower_schema, upper_schema),
            _ => continue, // We already picked up PKs from the lower sync
        }?;
        projection.push(expr);

        col_desc.insert(ColumnDescriptor {
            role: upper_sync_col.role() as _,
            name: upper_sync_col.name().clone(),
        });
    }

    let col_desc = col_desc.into_iter().collect::<Vec<_>>();

    // Finally use a qualified alias for the projection, which will act as a new lower sync in the
    // subsequent sync merge
    let projection = col_desc
        .iter()
        .zip(projection)
        .map(|(cd, expr)| {
            expr.alias_qualified(Some(LOWER_SYNC), SyncColumn::canonical_field_name(cd))
        })
        .collect();

    Ok((col_desc, projection))
```

src/frontend/flight/sync/utils.rs
gruuya: Yeah, I was looking into doing something like that, but there are some (non-)subtle differences between lower and upper sync column handling, which precluded them from sharing the same function. Or at least it would necessitate adding a bunch of extra parameters for instance the `WHEN` clause match expr, relation qualifier, the second `THEN` clause expr (the one in `otherwise`) etc.

In addition, the  `project_value_columns` would need to have `mut projection` and `mut col_desc` since there's a case when it needs to add an extra column that is missing in both upper and lower schema.

Let me see if I can simplify it somehow.

gruuya: Ok I refactored it a bit, should be a bit more readable now.

lizardoluis (approved):

src/frontend/flight/sync/utils.rs
gruuya: Ok, merged `project_lower_changed_column` and `project_upper_changed_column` into `project_changed_column`, I think this helps too.

Will try compressing Value projection too.

lizardoluis: With the lower and upper functions is already much more readable. Thank you for the changes

src/frontend/flight/sync/writer.rs
```diff
@@ -645,12 +638,110 @@ impl SeafowlDataSyncWriter {
...
 
-    fn apply_sync(
+    // Perform logical squashing of all pending sync batches into a single dataframe/plan, which can
+    // then be joined against the base scan
+    fn squash_syncs(
+        &self,
+        syncs: &[DataSyncItem],
+    ) -> SyncResult<(SyncSchema, DataFrame)> {
+        let mut sync_schema = syncs.first().unwrap().sync_schema.clone();
+        let first_batch = syncs.first().unwrap().batch.clone();
```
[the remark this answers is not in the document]
gruuya: Thanks, applied as well. To be clear the `first_sync`/`first_batch` is being unwraped only once per flush.

src/frontend/flight/sync/utils.rs
```diff
@@ -414,6 +415,288 @@ pub(super) fn get_prune_map(
...
+        col(UPPER_SYNC).is_null(),
+        col(Column::new(Some(LOWER_SYNC), lower_new_pk_field)),
+    )
+    .otherwise(col(Column::new(
+        Some(UPPER_SYNC),
+        upper_new_pk_col.field().name(),
+    )))?)
+}
+
+fn project_changed_column(
```
gruuya, line 550: I feel like the merged upper/lower routine now have too much stuff going on, and are actually detrimental to understanding, will split them up.

mildbyte (approved):

src/frontend/flight/sync/writer.rs
```diff
@@ -645,12 +638,113 @@ impl SeafowlDataSyncWriter {
...
+                    .field()
+                    .name()
+                    .clone();
+                let upper_name = sc.field().name().clone();
+                (lower_name, upper_name)
+            })
+            .into_iter()
+            .unzip();
+
+        // TODO merge syncs using a union if schemas match
```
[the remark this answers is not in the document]
gruuya: The comment is misleading, I meant go for a union if the schemas match _and_ the upper sync is a append-only batch (e.g. checked via `Array::null_count == num_rows` on a PK column).

PR #648, opened 2024-09-02
Index sync columns

gruuya:
To make the lookups during logical sync squashing faster pre-index the sync schema columns by role and name.

Also do `git mv src/frontend/flight/sync/* src/sync`.

### DIFF
--- a/src/sync/schema.rs
+++ b/src/sync/schema.rs
@@ -1,4 +1,4 @@
-use crate::frontend::flight::sync::SyncError;
+use crate::sync::SyncError;
 use arrow_schema::{DataType, FieldRef, SchemaRef};
 use clade::sync::{ColumnDescriptor, ColumnRole};
 use std::collections::HashSet;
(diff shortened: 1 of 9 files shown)


src/sync/schema.rs
```diff
@@ -1,11 +1,13 @@
...
 use arrow_schema::{DataType, FieldRef, SchemaRef};
 use clade::sync::{ColumnDescriptor, ColumnRole};
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
+use std::sync::Arc;
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct SyncSchema {
     columns: Vec<SyncColumn>,
+    indices: HashMap<ColumnRole, HashMap<Arc<str>, usize>>,
```
gruuya, line 10: This might be an overkill, since if the tables are wide, the baseline duration on flushing is also going to increase anyway.

An compromise solution I could just use `IndexMap` here instead? 
It would be less complex code, but we also get indexed column lookups (though not automatic filtering in `map_columns`).

gruuya: I've been testing this a bit on a 100-column table, and noticed that the speedup is modest, since the bulk of the logical sync squashing time is actually spent in a `DataFrame::with_column` call (fix incoming).

I still think we need indexing, but am leaning more on just using a single `IndexMap` to achieve this with minimal overhead.

mildbyte (approved):

PR #657, opened 2024-09-06
Fix flaky. Workaround for SQLite driver bug returning a result before the commit.

lizardoluis:
There is a bug on SQLite driver, which makes it return a result before the actual commit.

When using `fetch_one()` for queries with a RETURNING clause, the query prematurely returns after fetching the first result row without ensuring the SQLite statement is fully completed.

This bug was causing the test `test_sqlite_repository_read_only()` to fail sometimes when creating a database at https://github.com/splitgraph/seafowl/blob/8a47781f82ad54c445940f0d2c9e1008d6ddc2fa/src/repository/default.rs#L142-L143. 

In this PR, I implement a workaround by manually starting a transaction before running the query and committing it afterwards.

Flaky failure example: 

https://github.com/splitgraph/seafowl/actions/runs/10721496337/job/29730146388

### DIFF
--- a/src/repository/default.rs
+++ b/src/repository/default.rs
@@ -137,11 +137,15 @@ impl Repository for $repo {
     }
 
     async fn create_database(&self, database_name: &str) -> Result<DatabaseId, Error> {
+        let mut tx = self.executor.begin().await.map_err($repo::interpret_error)?;
+
         let id = sqlx::query(r#"INSERT INTO database (name) VALUES ($1) RETURNING (id)"#)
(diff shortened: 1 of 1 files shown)


gruuya (approved):
Great find!

For the sake of simplicity my only recommendation would be to use the tx only for DML using the `RETURNING` clause (the potentially buggy ones), and leave the rest of the queries as before (the pure read ones such as `get_database`).

lizardoluis:
Ok, I removed the tx from the others

PR #658, opened 2024-09-06
Extract the sync planning logic into a separate module

gruuya:
No (major) new things added here, just helping with code readability a bit.

The only extra thing is the planning time metric.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,10 @@ on: [push, pull_request]
 
 jobs:
   CI:
+    # Enable pull request trigger only from forks, since otherwise we get 2 job runs in PRs
+    if:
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -4,6 +4,10 @@ on: [push, pull_request]
 
 jobs:
   CI:
+    # Enable pull request trigger only from forks, since otherwise we get 2 job runs in PRs
+    if:
+      github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name !=
+      github.repository
```
gruuya, line 10: I seems it's either this or revising the `on` trigger to
```
on:
  push:
    branches:
      - master
  pull_request:
```
in order to avoid double CI runs (in the later case CI doesn't get run for pushed branches without an open PR).

mildbyte (approved):

PR #663, opened 2024-09-10
Improve Delta Table loading error message

mildbyte:
Inject context with the actual table name we were trying to load

Before:

```
External error: Not a Delta table: no log files
```

After:

```
loading Delta Table public.replicated_table
caused by
External error: Not a Delta table: no log files
```

### DIFF
--- a/src/provider.rs
+++ b/src/provider.rs
@@ -98,7 +98,12 @@ impl SchemaProvider for SeafowlSchema {
             },
         };
 
-        delta_table.load().await?;
+        delta_table.load().await.map_err(|e| {
+            DataFusionError::Context(
+                format!("loading Delta Table {}.{}", self.name, name),
+                Box::new(DataFusionError::External(Box::new(e))),
+            )
+        })?;
 
         let table = Arc::from(delta_table) as Arc<dyn TableProvider>;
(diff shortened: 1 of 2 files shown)


gruuya (approved):

src/provider.rs
```diff
@@ -98,7 +98,12 @@ impl SchemaProvider for SeafowlSchema {
             },
         };
 
-        delta_table.load().await?;
+        delta_table.load().await.map_err(|e| {
+            DataFusionError::Context(
+                format!("loading Delta Table {}.{}", self.name, name),
```
gruuya, line 103: Might be also useful to include the table url as well.

mildbyte: good idea fixed

```
Loading Delta Table public.replicated_table (memory:///61b8bdab-45dd-479b-a7d5-d56dde4e56f7)
caused by
External error: Not a Delta table: no log files
```

eatonphil: Also I did not realize we could use memory tables!

mildbyte: Yep, you can create an in-memory storage location and write a table there, but it will go away on Seafowl restart :)

PR #666, opened 2024-09-11
Make replication idempotent against commited rows

gruuya:
## What

To make replication robust in the face of failures and crashes we need to be able to apply the same messages multiple times without the state deviating from the expected one. This should hold in general, i.e. for INSERTs, DELETEs, and UPDATEs (whether they change the PK or not).

## How

Split up PK-changing UPDATEs into DELETEs + INSERTs, which makes them idempotent, assuming all columns (TOASTed or not) are supplied for such cases (since otherwise the INSERT has no idea where to pull the old values from).

## Why

Ultimately this is the fastest method method to get only-once semantics working, with alternatives such as aggregation/grouping/distinct-on etc. being much slower.

### DIFF
--- a/src/frontend/flight/handler.rs
+++ b/src/frontend/flight/handler.rs
@@ -24,8 +24,6 @@ use crate::sync::schema::SyncSchema;
 use crate::sync::writer::SeafowlDataSyncWriter;
 use crate::sync::SyncResult;
 
-pub const SEAFOWL_SYNC_CALL_MAX_ROWS: usize = 65536;
-
 lazy_static! {
     pub static ref SEAFOWL_SQL_DATA: SqlInfoData = {
         let mut builder = SqlInfoDataBuilder::new();
--- a/src/sync/writer.rs
+++ b/src/sync/writer.rs
@@ -257,7 +257,7 @@ impl SeafowlDataSyncWriter {
             self.metrics.squash_time.record(duration as f64);
             self.metrics
                 .squashed_bytes
-                .increment((sync_size - size) as u64);
+                .increment((sync_size.saturating_sub(size)) as u64);
             self.metrics
                 .squashed_rows
                 .increment((sync_rows - rows) as u64);
@@ -624,7 +624,7 @@ mod tests {
     use crate::sync::schema::{arrow_to_sync_schema, SyncSchema};
     use crate::sync::writer::{SeafowlDataSyncWriter, SequenceNumber};
     use crate::sync::{SyncCommitInfo, SyncResult};
-    use arrow::array::{BooleanArray, Float32Array, Int32Array};
+    use arrow::array::{BooleanArray, Float32Array, Int32Array, StringArray};
     use arrow::{array::RecordBatch, util::data_gen::create_random_batch};
     use arrow_schema::{DataType, Field, Schema, SchemaRef};
     use datafusion_common::assert_batches_eq;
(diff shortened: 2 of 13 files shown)


src/sync/writer.rs
```diff
@@ -1080,29 +1080,34 @@ mod tests {
...
     #[rstest]
-    #[case(vec![(vec![1, 2], vec![1, 2])])]
-    #[case(vec![(vec![1, 2], vec![2, 1]), (vec![1, 2], vec![2, 1])])]
+    #[case(vec![(vec![1, 2], vec![1, 2], None)])]
+    #[should_panic(expected = "assertion `left == right` failed")]
+    #[case(vec![
+        (vec![1, 2], vec![2, 1], Some(vec!["a", "b"])),
+        (vec![2, 1], vec![1, 2], Some(vec!["a", "b"]))
+    ])]
     #[case(vec![
```
gruuya, line 1092: Turns out this is an edge case i missed; what happens is that physical squashing compresses
```rust
vec![1, 2], vec![2, 1], Some(vec!["a", "b"])
```
into
```rust
1, 1, b
```
and it compresses 
```rust
vec![2, 1], vec![1, 2], Some(vec!["a", "b"])
```
into 
```rust
2, 2, b
```
since it detects a PK-chain.

Consequently, the post-apply state ends up being
```rust
1, b
2, b
```
which is wrong.

This necessitates the we normalize the batch prior to physical squashing, or ideally just alter the protocol to how it was before (as it is resiliant against this edge case). Going back to the previous protocol version will also simplify a lot of the planning logic too.

gruuya: Actually, thinking about this some more, this seems to be an impossible input batch for the writer.

This is because it violates the PK-conflict rules, which is an invariant that I'm working with (i.e. upstream is responsible for ensuring the original PK rules are enforced and reflected in the change batches).

So what needs to be done is for the test case to get updated along those lines.

mildbyte: Would this be fine if the test case was 2->3, 1->2, 3->1 (i.e. we swap the PKs using a third temporary one)? What if the user temporarily suspends the PK constraint during a transaction (or it's deferred to the end of the TX) https://www.postgresql.org/docs/current/sql-set-constraints.html and is still valid when the TX commits)?

src/sync/writer.rs
```diff
@@ -946,13 +950,49 @@ mod tests {
...
+        1000,
+        1999,
+    )]
+    // #[case::insert_update_idempotent(
+    //     vec![None, Some(0..300), Some(0..300), Some(300..1000), Some(300..1000)],
+    //     vec![Some(0..1000), Some(1000..1300), Some(1000..1300), Some(1300..2000), Some(1300..2000)],
+    //     1000,
+    //     1000,
+    //     1999,
+    // )]
```
mildbyte, line 980: What happened to the commented out cases, do they need to be updated/fixed?

mildbyte (approved):

src/sync/writer.rs
gruuya: They're the ones where in-memory idempotence is required, so for now I'd leave them commented out.

src/sync/writer.rs
gruuya: > Would this be fine if the test case was 2->3, 1->2, 3->1 (i.e. we swap the PKs using a third temporary one)?

Yes, the replacement test case does this https://github.com/splitgraph/seafowl/blob/b0efbd3306d30b96c3dd97ad40d606f27adb92a4/src/sync/writer.rs#L1098-L1101

> What if the user temporarily suspends the PK constraint during a transaction (or it's deferred to the end of the TX) https://www.postgresql.org/docs/current/sql-set-constraints.html and is still valid when the TX commits)?

We're vulnerable against that until we switch over to the upsert batch format.

src/frontend/flight/handler.rs
```diff
@@ -24,8 +24,6 @@ use crate::sync::schema::SyncSchema;
 use crate::sync::writer::SeafowlDataSyncWriter;
 use crate::sync::SyncResult;
 
-pub const SEAFOWL_SYNC_CALL_MAX_ROWS: usize = 65536;
```
eatonphil, line 27: Ah ok so there's no max now at all is that right?

gruuya: Yeah, that is correct. Though keep in mind that there is a slight throughput regression the more the batches grow beyond ~100K rows.

PR #683, opened 2024-09-27
Fix prefix swallowing when creating the object store

gruuya:
Append the table path to the pre-existing url path instead of replacing it.

### DIFF
--- a/src/object_store/factory.rs
+++ b/src/object_store/factory.rs
@@ -99,7 +99,7 @@ impl ObjectStoreFactory {
     }
     pub async fn get_log_store_for_table(
         &self,
-        url: Url,
+        mut url: Url,
         options: HashMap<String, String>,
         table_path: String,
     ) -> Result<Arc<dyn LogStore>, object_store::Error> {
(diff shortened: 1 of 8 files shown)


src/object_store/factory.rs
```diff
@@ -134,16 +134,28 @@ impl ObjectStoreFactory {
...
+            && url.scheme() != "memory"
+        {
+            format!("{}/{table_path}", url.path())
+        } else {
+            table_path.clone()
+        };
+
+        // This is the least surprising way to extend the path, and make the url point to the table
+        // root: https://github.com/servo/rust-url/issues/333
+        url.path_segments_mut().unwrap().push(&table_path);
```
gruuya, line 151: `Url::join` is unreliable/surprising, so it actually will swallow up the prefix, whereas this works.

mildbyte: can this unwrap() be easily replaced by an error?

mildbyte (approved):

src/object_store/factory.rs
gruuya: Yup, replaced.

PR #690, opened 2024-10-01
Sync batch coalescing

gruuya:
## What

Coalesce sync batches when enqueuing them. This ensures we stay below the `max_syncs_per_url` for longer, ultimately enabling us to accumulate/dump more rows per flush.

## How

When enqueuing sync batches, if they share the same schema as a previous entry just append them to it, instead of creating a new entry.

## Why

Whereas the existing implementation has been optimized for a high volume/saturating use case, this is not the case for the opposite, i.e. batches with a low number of rows. Consequently this leads to poor throughput rates, and ultimately slows down the transaction processing speed for the entire cluster.

### DIFF
--- a/src/sync/schema/mod.rs
+++ b/src/sync/schema/mod.rs
@@ -152,9 +152,9 @@ impl SyncSchema {
 
     // Check whether this and the other sync schemas are the same.
     //
-    // This is a narrower check than the equality check, since it checks for the sync column roles,
+    // This is a shallower check than the equality check, since it checks for the sync column roles,
     // names, data types and order (and doesn't inspect the underlying arrows field equality).
-    pub fn same(&self, other: &SyncSchema) -> bool {
+    pub fn is_compatible_with(&self, other: &SyncSchema) -> bool {
         self.columns()
             .iter()
             .zip(other.columns().iter())
--- a/src/sync/schema/mod.rs
+++ b/src/sync/schema/mod.rs
@@ -149,6 +149,21 @@ impl SyncSchema {
             })
             .unwrap_or_default()
     }
+
+    // Check whether this and the other sync schemas are the same.
+    //
+    // This is a narrower check than the equality check, since it checks for the sync column roles,
+    // names, data types and order (and doesn't inspect the underlying arrows field equality).
+    pub fn same(&self, other: &SyncSchema) -> bool {
+        self.columns()
+            .iter()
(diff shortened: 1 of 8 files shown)


src/sync/schema/mod.rs
```diff
@@ -149,6 +149,21 @@ impl SyncSchema {
             })
             .unwrap_or_default()
     }
+
+    // Check whether this and the other sync schemas are the same.
+    //
+    // This is a narrower check than the equality check, since it checks for the sync column roles,
+    // names, data types and order (and doesn't inspect the underlying arrows field equality).
+    pub fn same(&self, other: &SyncSchema) -> bool {
```
lizardoluis, line 157:
```suggestion
    pub fn is_compatible_with(&self, other: &SyncSchema) -> bool {
```

or `has_same_sync_schema`, `matches_sync_columns`

src/sync/utils.rs
```diff
@@ -13,18 +13,28 @@ use datafusion_expr::{col, lit, Accumulator, Expr};
 use std::collections::{HashMap, HashSet, VecDeque};
 use tracing::log::warn;
 
+// Returns the total number of bytes and rows in the slice of batches
+pub(super) fn size_and_rows(batches: &[RecordBatch]) -> (usize, usize) {
```
lizardoluis, line 17:
```suggestion
pub(super) fn compute_size_and_rows(batches: &[RecordBatch]) -> (usize, usize) {
```
or `get_size_and_rows`

src/sync/writer.rs
```diff
@@ -113,13 +115,16 @@ pub(super) struct DataSyncCollection {
 // An object corresponding to a single `do_put` call.
 #[derive(Debug, Clone, PartialEq)]
 pub(super) struct DataSyncItem {
+    // Flag denoting whether the batches have been physically squashed yet or not. If true there is
+    // a single record batch in the `data` field, and this is strictly the case during flushing.
+    pub(super) squashed: bool,
```
lizardoluis, line 120:
```suggestion
    pub(super) has_been_squashed: bool,
```
or `is_squashed`. Because then it shows that the batches have been changed, and not is a flag allowing it.

lizardoluis (approved):
My comments were all about naming to improve readability. It looks good to me, but I think we need unittests on all these functions, giving the complexity of these algorithms.

mildbyte (approved):

PR #704, opened 2024-10-14
Handle prefixes for AmazonS3 and GoogleCloudStorage object stores

SergeiPatiakin:
Add prefixing logic for S3 and GCP object stores in object_store_factory::build_object_store_from_opts. This makes prefixing consistent across all object store types.

### DIFF
--- a/object_store_factory/src/lib.rs
+++ b/object_store_factory/src/lib.rs
@@ -126,14 +126,20 @@ pub async fn build_object_store_from_opts(
             aws::add_amazon_s3_specific_options(url, &mut s3_options).await;
             aws::add_amazon_s3_environment_variables(&mut s3_options);
 
-            let (store, _) = parse_url_opts(url, s3_options)?;
+            let (mut store, _) = parse_url_opts(url, s3_options)?;
+            if !url.path().is_empty() {
+                store = Box::new(PrefixStore::new(store, url.path()));
+            }
             Ok(store)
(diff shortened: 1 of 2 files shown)


object_store_factory/src/lib.rs
```diff
@@ -126,14 +126,20 @@ pub async fn build_object_store_from_opts(
             aws::add_amazon_s3_specific_options(url, &mut s3_options).await;
             aws::add_amazon_s3_environment_variables(&mut s3_options);
 
-            let (store, _) = parse_url_opts(url, s3_options)?;
+            let (mut store, _) = parse_url_opts(url, s3_options)?;
+            if !url.path().is_empty() {
+                store = Box::new(PrefixStore::new(store, url.path()));
+            }
```
gruuya, line 132: We should be able to just use the path returned by the object_store crate here and below.
```suggestion
            let (mut store, path) = parse_url_opts(url, s3_options)?;
            if !path.is_empty() {
                store = Box::new(PrefixStore::new(store, path));
            }
```

gruuya (approved):

object_store_factory/src/lib.rs
SergeiPatiakin: That path is of type `object_store::path::Path` which doesn't have an `is_empty` method. I would stick to url.path() as it's what we use further up in L115

PR #707, opened 2024-10-17
Add an option to the local FS config to make all missing directories on the path

gruuya:
Also use that option in the replication interface (otherwise the object store creation may fail).

There is an alternative simpler change, which is also less flexible, and that is to simply do `LocalFileSystem::new()` in `LocalConfig::build_local_storage` and then wrap that in a `PrefixStore`. I don't particularly like that option due to the redundant wrapping, and for the fact that it doesn't do validation (which we may want in some cases).

### DIFF
--- a/object_store_factory/src/local.rs
+++ b/object_store_factory/src/local.rs
@@ -1,13 +1,19 @@
 use object_store::{local::LocalFileSystem, ObjectStore};
 use serde::Deserialize;
 use std::collections::HashMap;
+use std::fs;
 use std::sync::Arc;
+
+pub const DATA_DIR: &str = "data_dir";
+pub const CREATE_DIR: &str = "create_dir";
 
 #[derive(Deserialize, Debug, PartialEq, Eq, Clone)]
 pub struct LocalConfig {
     pub data_dir: String,
     #[serde(default = "default_false")]
     pub disable_hardlinks: bool,
+    #[serde(default = "default_false")]
+    pub create_dir: bool,
 }
 
 fn default_false() -> bool {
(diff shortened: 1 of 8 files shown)


object_store_factory/src/local.rs
```diff
@@ -46,6 +53,16 @@ impl LocalConfig {
     pub fn build_local_storage(
         &self,
     ) -> Result<Arc<dyn ObjectStore>, object_store::Error> {
+        if self.create_dir {
+            // `LocalFileSystem::new_with_prefix` will error out unless the path fully exists
+            fs::create_dir_all(self.data_dir.clone()).map_err(|e| {
```
mildbyte, line 58: Does this error if the directory already exists? i.e. if we have two writes, will the second one break?

lizardoluis:
@gruuya can you please port this to the [PGFS](https://github.com/EnterpriseDB/pgfs) as well?

src/frontend/flight/handler.rs
```diff
@@ -168,6 +169,12 @@ impl SeafowlFlightHandler {
...
             Some(store_loc) => {
+                let mut options = store_loc.options;
+                if store_loc.location.starts_with("file://") {
+                    // In case of a local FS create the table path
+                    options.insert(CREATE_DIR.to_string(), "true".to_string());
+                }
+
                 self.context
                     .metastore
                     .object_stores
```
mildbyte, line 180: Come to think about it, maybe I'd add it here instead, add a parameter to this function like `get_log_store_for_table(for_writes: bool)` and then avoid having to thread through a magic CREATE_DIR setting and figure out which location it is based on the protocol which feels like a leaky abstraction.

gruuya: Hmm, doing that kinda feels like a leaky abstraction too? 

I would either need to pass `for_writes: bool` down into `object_store_factory::build_object_store_from_opts` too (which seems redundant, all object stores should be writable), or add CREATE_DIR at that level to the optoins based on the protocol/scheme, same as now.

Maybe the best solution is just to go with wrapping `LocalFileSystem::new()` with an appropriate `PrefixStore` then?

gruuya:
Closing in favor of #708

PR #717, opened 2024-10-22
Align batch schema nullability during squashing

gruuya:
This prevents arrow `Column 'x' is declared as non-nullable but contains null values` errors during record batch creation.

### DIFF
--- a/src/sync/schema/mod.rs
+++ b/src/sync/schema/mod.rs
@@ -28,7 +28,7 @@ impl SyncSchema {
         }
 
         // Validate field role's are parsable, we have the correct number of old/new PKs,
-        // and Changed role is non-nullable boolean type which points to an existing column
+        // and Changed role is a boolean type which points to an existing column
         // TODO: Validate a column can not be a PK and Value at the same time
         let mut old_pk_types = HashSet::new();
         let mut new_pk_types = HashSet::new();
--- a/src/sync/schema/mod.rs
+++ b/src/sync/schema/mod.rs
@@ -126,17 +126,18 @@ impl SyncSchema {
     pub fn with_fields(&mut self, fields: &Fields) -> SyncResult<()> {
         if fields.len() != self.columns.len() {
             return Err(SyncError::SchemaError {
-                reason: "New and old field count is different".to_string(),
+                reason: "New and old field counts are different".to_string(),
             });
         }
 
         for (col, field) in self.columns.iter_mut().zip(fields.iter()) {
-            if col.field.data_type() != field.data_type() {
+            let col_data_type = col.field.data_type();
+            let field_data_type = field.data_type();
+
+            if col_data_type != field_data_type {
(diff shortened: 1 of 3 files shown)


src/sync/utils.rs
```diff
@@ -29,12 +31,30 @@ pub(super) fn get_size_and_rows(batches: &[RecordBatch]) -> (usize, usize) {
...
 pub(super) fn squash_batches(
-    sync_schema: &SyncSchema,
+    sync_schema: &mut SyncSchema,
     data: &[RecordBatch],
-) -> Result<RecordBatch> {
-    // Concatenate all the record batches into a single one
-    let schema = data.first().unwrap().schema();
+) -> SyncResult<RecordBatch> {
     debug!("Concatenating {} batch(es)", data.len());
+    // Concatenate all the record batches into a single one, after making all fields nullable first.
```
gruuya, line 38: There are several alternatives to this, since data types have already been validated, and we really only want to align on nullability:

1. Add a validation step in `SyncSchema::try_new` that rejects messages where a field is not nullable
2. Align the schemas with `let schema = Arc::new(Schema::try_merge(data.iter().map(|batch| batch.schema().as_ref().clone()))?)`

The later is a more costly way to basically achieve the same thing as here (as it needs to iterate/clone all the schemas and iterate through all fields checking each one individually).

The former is a much simpler solution/implementation, though slightly reducing flexibility.

src/sync/schema/mod.rs
```diff
@@ -122,6 +122,31 @@ impl SyncSchema {
         }
     }
 
+    // Replace the existing field references for sync columns with new ones.
+    pub fn with_fields(&mut self, fields: &Fields) -> SyncResult<()> {
+        if fields.len() != self.columns.len() {
+            return Err(SyncError::SchemaError {
+                reason: "New and old field count is different".to_string(),
```
lizardoluis, line 129:
```suggestion
                reason: "New and old field count are different".to_string(),
```

src/sync/schema/mod.rs
```diff
@@ -122,6 +122,31 @@ impl SyncSchema {
...
+                    reason: format!(
+                        "Expected {} but got {} for column {col:?}",
+                        col.field.data_type(),
+                        field.data_type()
+                    ),
+                });
+            }
+
+            col.field = field.clone();
+        }
```
lizardoluis, line 145:
```suggestion
        for (col, field) in self.columns.iter_mut().zip(fields.iter()) {
            let col_data_type = col.field.data_type();
            let field_data_type = field.data_type();
            
            if col_data_type != field_data_type {
                return Err(SyncError::SchemaError {
                    reason: format!(
                        "Schema mismatch for column {col:?}: expected data type {}, but got {}",
                        col_data_type,
                        field_data_type
                    ),
                });
            }

            col.field = field.clone();
        }
```

lizardoluis (approved):

PR #727, opened 2024-11-05
Add support for memory profiling in MacOS

gruuya:
This uses a dockerized bytehound profiler and a minimal justfile to record and view the profiles.


SergeiPatiakin (approved):

seafowl.toml
```diff
@@ -15,4 +15,4 @@ bind_port = 6432
 bind_host = "0.0.0.0"
 
 [frontend.flight]
-bind_host = "127.0.0.1"
+bind_host = "0.0.0.0"
```
mildbyte, line 18: Can you do it as an envvar in the Dockerfile instead? Seeing 0.0.0.0 anywhere, even in a default dev config file, gets me worried :D

mildbyte (approved):

PR #743, opened 2024-11-21
Add connection Credentials to StorageLocation

lizardoluis:
…tionality


object_store_factory/src/google.rs
```diff
@@ -245,18 +261,21 @@ mod tests {
 
     #[test]
     fn test_map_options_into_google_config_keys_with_valid_keys() {
-        let mut input_options = HashMap::new();
-        input_options.insert(
+        let mut options = HashMap::new();
+        options.insert(
             "google_service_account".to_string(),
             "my_google_service_account".to_string(),
         );
```
gruuya, line 268: Shouldn't this be in `credentials` now?

gruuya (commented):
I'm not sure I follow why we need to separate out the credentials in the first place? In this way we add some friction to the API, with one having to delineate what is an option and what is a credential (whereas downstream crates, such as `object_store` and [iceberg rust storage](https://github.com/apache/iceberg-rust/blob/697a20060f2247da87f73073e8bf5ab407bd40ea/crates/iceberg/src/io/storage_s3.rs#L65-L82) treat them the same).

Since they end up in the same place anyway why just not simply merge the two maps prior to calling `build_object_store_from_opts`/`S3config::from_hashmap` and leave the `object_store_factory` as is?

PR #751, opened 2024-11-27
Add table format parameter to clade

gruuya:
Also adds Iceberg table instantiation, and a corresponding integration test.


Cargo.toml
```diff
@@ -100,6 +100,7 @@ deltalake = { git = "https://github.com/splitgraph/delta-rs", branch = "fix-deci
 
 futures = "0.3"
 hex = ">=0.4.0"
+iceberg = { git = "https://github.com/apache/iceberg-rust", rev = "697a20060f2247da87f73073e8bf5ab407bd40ea" }
```
SergeiPatiakin, line 103: On main we changed the iceberg-datafusion repo to `splitgraph/iceberg-rust`. I suggest updating this branch from main and also using `splitgraph/iceberg-rust` for this new dependency

src/object_store/utils.rs
```diff
@@ -44,3 +48,33 @@ pub async fn fast_upload(from: &StdPath, to: String) -> object_store::Result<(),
...
+            _ => key, // for now just propagate any non-matched keys
+        };
+
+        props.insert(key.to_string(), val.clone());
+    }
+
+    // FileIO requires the region prop even when the S3 store doesn't (e.g. MinIO)
+    props
+        .entry(S3_REGION.to_string())
+        .or_insert("us-east-1".to_string());
```
SergeiPatiakin, line 77: Do you think we should use a dummy value like "dummy" so that it fails when using real AWS S3 (rather than using real AWS S3 with an incorrect region)?

gruuya: That's a good point, doing that.

Cargo.toml
gruuya: Rebased and pointed to our fork.

SergeiPatiakin (approved):

PR #753, opened 2024-11-27
Prepare for Iceberg support

SergeiPatiakin:
- Add enum LakehouseTableProvider that has variants for delta and iceberg
- Add method SeafowlContext::get_lakehouse_table_provider. To add iceberg support, methods from `physical.rs` in future PRs can:
	- call this instead of SeafowlContext::try_get_delta_table
	- implement separate logic for delta and iceberg variants of the table provider

### DIFF
--- a/src/context/delta.rs
+++ b/src/context/delta.rs
@@ -89,7 +89,7 @@ fn temp_partition_file_writer(
 
 /// Execute a plan and upload the results to object storage as Parquet files, indexing them.
 /// Partially taken from DataFusion's plan_to_parquet with some additions (file stats, using a DiskManager)
-pub async fn plan_to_delta_data(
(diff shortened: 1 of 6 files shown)


src/context/delta.rs
```diff
@@ -89,7 +89,7 @@ fn temp_partition_file_writer(
 
 /// Execute a plan and upload the results to object storage as Parquet files, indexing them.
 /// Partially taken from DataFusion's plan_to_parquet with some additions (file stats, using a DiskManager)
-pub async fn plan_to_object_store(
+pub async fn plan_to_delta_data(
```
gruuya, line 92: Maybe also
```suggestion
pub async fn plan_to_delta_adds(
```
to make it more concrete

gruuya (approved):

PR #758, opened 2024-12-09
Move FileIO props conversion into the object_store_factory

gruuya:
This makes it useful beyond just Seafowl.

### DIFF
--- a/object_store_factory/src/utils.rs
+++ b/object_store_factory/src/utils.rs
@@ -1,0 +1,45 @@
+use iceberg::io::{
+    S3_ACCESS_KEY_ID, S3_ALLOW_ANONYMOUS, S3_DISABLE_EC2_METADATA, S3_ENDPOINT,
+    S3_REGION, S3_SECRET_ACCESS_KEY,
+};
+use object_store::aws::AmazonS3ConfigKey;
+use std::collections::HashMap;
+use std::str::FromStr;
+
+// Go through all known keys for object store and convert them to corresponding file_io ones.
+//
+// For now only converts S3 keys.
+// TODO: At some point this should be redundant, since there is an OpenDAL adapter for object_store,
+// https://github.com/apache/iceberg-rust/issues/172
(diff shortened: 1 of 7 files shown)


SergeiPatiakin (approved):

object_store_factory/src/utils.rs
```diff
@@ -0,0 +1,45 @@
...
+use object_store::aws::AmazonS3ConfigKey;
+use std::collections::HashMap;
+use std::str::FromStr;
+
+// Go through all known keys for object store and convert them to corresponding file_io ones.
+//
+// For now only converts S3 keys.
+// TODO: At some point this should be redundant, since there is an OpenDAL adapter for object_store,
+// https://github.com/apache/iceberg-rust/issues/172
+pub fn object_store_opts_to_file_io_props(
```
lizardoluis, line 14: @gruuya sorry for the after merge review, you guys were faster than me.... but ..

This function should be implemented at `object_store_factory/src/aws.rs`, together with every other that deals with `aws`. 

And following the pattern and thinking about extensibility,  we need to add one at `object_store_factory/src/lib.rs` that calls it if it is for `aws`. Just like the others.

object_store_factory/src/utils.rs
```diff
@@ -0,0 +1,45 @@
...
+    opts: &HashMap<String, String>,
+) -> HashMap<String, String> {
+    let mut props = HashMap::new();
+
+    for (key, val) in opts.iter() {
```
lizardoluis, line 19: We are using a `HashMap<String, String>` for `options` throughout the code, but I found out another day that the `HashMap` is not 'that' efficient for iterations, as it iterates through the capacity and not the size. 

If the caller of this function does not need a HashMap for it, I suggest to pass a List or Vec of options here instead of the HashMap

object_store_factory/src/utils.rs
lizardoluis: You can put it then on lib.rs

PR #759, opened 2024-12-09
Iceberg write support

SergeiPatiakin:
With this PR, a user can use the DML `INSERT` statement to append to an existing Iceberg table.

Iceberg code is borrowed from https://github.com/splitgraph/lakehouse-loader . It makes some accommodation for creating Iceberg tables from scratch, which is not supported in this PR but will be supported in future.


tests/data/iceberg/default.db/iceberg_table_2/README.md
```diff
@@ -0,0 +1,32 @@
+This table was generated with the following code:
```
gruuya, line 1: Could also mention that we manually renamed the metadata version files to comply with the FS-based convention.

src/context/iceberg.rs
```diff
@@ -0,0 +1,379 @@
...
+
+impl SeafowlContext {
+    pub async fn plan_to_iceberg_table(
+        &self,
+        name: impl Into<TableReference>,
+        plan: &Arc<dyn ExecutionPlan>,
+    ) -> Result<()> {
+        let provider = match self.get_lakehouse_table_provider(name).await? {
+            LakehouseTableProvider::Iceberg(p) => p,
+            _ => panic!("Expected iceberg provider"),
```
gruuya, line 357: We should probably return an error here instead of panicking.

src/context/iceberg.rs
```diff
@@ -0,0 +1,379 @@
...
+        };
+        let table = provider.table();
+
+        for i in 0..plan.output_partitioning().partition_count() {
+            let task_ctx = Arc::new(TaskContext::from(&self.inner.state()));
+            let stream = plan.execute(i, task_ctx)?;
```
gruuya, line 363: This suggests that we'll end up with a commit/new version of the table for each partition in the input plan, which is odd.

Can you merge the streams (e.g. into a `Vec<SendableRecordBatchStream>`, or some wrapper stream or perhaps by wrapping the plan with DataFusion `CoalescePartitionsExec`?) and then just call `record_batches_to_iceberg` once with the whole thing?

SergeiPatiakin: Good point. Will do `select_all` to merge streams

gruuya (approved):
Very nice, thanks!

PR #764, opened 2024-12-16
Add format field to DataSyncCommand

SergeiPatiakin:
Prepare for Iceberg sync support:

- Add table format field to DataSyncCommand (can be Delta or Iceberg)
- For now, require the format to be Delta

### DIFF
--- a/src/frontend/flight/handler.rs
+++ b/src/frontend/flight/handler.rs
@@ -4,6 +4,7 @@ use arrow::record_batch::RecordBatch;
 use arrow_flight::sql::metadata::{SqlInfoData, SqlInfoDataBuilder};
 use arrow_flight::sql::{ProstMessageExt, SqlInfo, TicketStatementQuery};
 use arrow_flight::{FlightDescriptor, FlightEndpoint, FlightInfo, Ticket};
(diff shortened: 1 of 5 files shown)


src/frontend/flight/handler.rs
```diff
@@ -162,6 +163,10 @@ impl SeafowlFlightHandler {
             });
         }
 
+        if cmd.format != TableFormat::Delta as i32 {
+            return Err(SyncError::NotImplemented);
```
gruuya, line 167: Maybe we can just use `SyncError::InvalidMessage` here since it's going to be temporary anyway.

gruuya (approved):

src/frontend/flight/handler.rs
SergeiPatiakin: I'm inclined to keep it. This functionality being unimplemented is temporary, but having some functionality unimplemented is permanent :)

"Not implemented" exists as a distinct concept in some languages e.g. Python: https://docs.python.org/3/library/constants.html#NotImplemented

gruuya: Fair point, we may need it in future.

PR #766, opened 2024-12-18
Prepare for Iceberg Sync

SergeiPatiakin:
This PR generalizes sync-related code from Delta-specific types towards a more abstract "sync target" type which can be Delta or Iceberg. We do not yet implement Iceberg sync functionality, that will be in a future PR.

### DIFF
--- a/src/frontend/flight/handler.rs
+++ b/src/frontend/flight/handler.rs
@@ -23,7 +23,7 @@ use url::Url;
 use crate::context::SeafowlContext;
 use crate::sync::schema::SyncSchema;
 use crate::sync::writer::SeafowlDataSyncWriter;
-use crate::sync::{SyncError, SyncResult};
+use crate::sync::{LakehouseSyncTarget, SyncError, SyncResult};
 
 lazy_static! {
     pub static ref SEAFOWL_SQL_DATA: SqlInfoData = {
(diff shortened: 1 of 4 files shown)


src/sync/mod.rs
```diff
@@ -56,6 +57,17 @@ pub(super) struct SyncCommitInfo {
     new_tx: bool,
 }
 
+#[derive(Clone, Debug)]
+pub struct IcebergSyncTarget {
+    url: String,
+}
+
+#[derive(Clone, Debug)]
+pub enum LakehouseSyncTarget {
```
SergeiPatiakin, line 66: Please share if you have a better name for this enum

src/sync/mod.rs
```diff
@@ -56,6 +57,17 @@ pub(super) struct SyncCommitInfo {
     new_tx: bool,
 }
 
+#[derive(Clone, Debug)]
+pub struct IcebergSyncTarget {
+    url: String,
```
SergeiPatiakin, line 62: This will have more fields in future

src/frontend/flight/handler.rs
```diff
@@ -163,32 +163,38 @@ impl SeafowlFlightHandler {
...
-                            ))
-                        })?,
-                        store_loc.options,
-                        cmd.path,
-                    )
-                    .await?
-            }
+        } else {
+            return Err(SyncError::InvalidMessage {
+                reason: "Unknown table format".to_string(),
```
gruuya, line 194: I think you might be able to match exactly on `cmd.format()` and avoid this part.

gruuya (approved):

PR #771, opened 2024-12-31
Iceberg sync

SergeiPatiakin:
Support Iceberg tables as sync targets


src/frontend/flight/handler.rs
```diff
@@ -186,14 +192,49 @@ impl SeafowlFlightHandler {
...
+                    location.trim_end_matches("/"),
+                    cmd.path.trim_start_matches("/")
+                );
+                let iceberg_table = StaticTable::from_metadata_file(
+                    &absolute_path,
+                    TableIdent::from_strs(vec![DEFAULT_SCHEMA, "dummy_name"]).unwrap(),
+                    file_io,
+                )
+                .await?
+                .into_table();
```
gruuya, line 227: I think this (or the line below) will only work if the table already exists at the passed location, whereas we support creating a brand new table from scratch for Delta.

Maybe `LakehouseSyncTarget::Iceberg` should only keep hold of `FileIO`, and then in `flush_syncs` we check whether a table exists, and if not create one first similar to here https://github.com/splitgraph/seafowl/blob/54bc3046e7f728a1fa9a30a025022c1ee7d6ee79/src/sync/writer.rs#L414-L422

tests/flight/sync.rs
```diff
@@ -609,3 +609,70 @@ async fn test_sync_custom_store(
 
     Ok(())
 }
+
+#[tokio::test]
+async fn test_sync_iceberg_custom_store(
```
gruuya, line 614: Seems like `test_sync_custom_store` and this test share the setup/results, maybe we can just parametrize the former to do exercise this routine for both formats (we can ditch the `local` parametrization in it since it won't work for Iceberg)?

src/sync/planner.rs
```diff
@@ -48,6 +48,31 @@ impl SeafowlSyncPlanner {
...
+        table_schema: Arc<arrow_schema::Schema>,
+        table_provider: Arc<dyn TableProvider>,
+    ) -> SyncResult<Arc<dyn ExecutionPlan>> {
+        // Convert the custom Iceberg table provider into a base logical plan
+        let base_plan = LogicalPlanBuilder::scan(
+            LOWER_REL,
+            provider_as_source(table_provider),
+            None,
+        )?
+        .build()?;
```
gruuya, line 63: The base scan/plan should only touch stuff that we're changing. In case of Iceberg we support only INSERTs for the time being, and so the base plan should be empty. Otherwise I think this will lead to data duplication on subsequent invocations, since we take everything from the current table state add the new stuff on top of it and dump that to a new table state (where we also inherit the manifests from the previous state)?

Also we might want to do some validation on batches having no UPDATES/DELETES, as those won't be correctly represented?

gruuya (commented):
Very nice, especially seeing how it is a smaller change than I anticipated.

tests/flight/sync.rs
SergeiPatiakin: The setup, verification and parametrization are all subtly different so I'm not sure it's worth it.

gruuya (approved):

